PR #4, opened 2020-11-30
Update galaxy.yml

gundalow:
##### SUMMARY
@tima 1) Do you have a better `description` to add? 2) Any other tags that should be added?


geerlingguy:
@gundalow - Note that for this collection, it is still in 'skeleton' mode until we move the content from `community.kubernetes` over here.

Issue #19, opened 2021-02-15
Implement "loop flattening" in `k8s`


tima:
##### SUMMARY

Using a loop to process a batch of file means a fork and API call is made for each item in the loop. This adds a substantial amount of overhead to process each file that could be avoided if the list of files are "flattened" and submitted with one API call. The proposal here is to build this "flattening" logic into the `k8s` module itself where optimal handling of a batch of files is trivial.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

- `k8s`

##### ADDITIONAL INFORMATION

```
- name: Create the Monitoring Dashboards
  k8s:
    state: "present"
    namespace: "{{ kiali_vars.deployment.namespace }}"
    definition: "{{ lookup('template', item) }}"
  loop: "{{ monitoring_dashboard_yamls_to_deploy }}"
  when:
    ...
```

While functional, using a loop to process each file means a fork and API call is also made for each item. This adds a substantial amount of overhead to process each file that could be avoided if the files are "flattened" in one pass and submitted with one API call.

Using creative use of Jinja2 syntax this can be done...

```
- name: Create the Monitoring Dashboards
  k8s:
    state: "present"
    namespace: "{{ kiali_vars.deployment.namespace }}"
    definition: |
      {% for mdy in monitoring_dashboard_yamls_to_deploy %}
      ---
      {{ lookup("template", mdy) }}
      ...
      {% endfor %}
  when:
    ...
```

This executes significantly faster than the prior example, but it also requires a fair bit of knowledge and effort to get right.

The proposal here is to build this "flattening" logic into the `k8s` module itself where optimal handling of batch file processing is trivial. 

The `src` and `template` (and by extension `template/path`) would be able to optionally accept a list of strings (file paths) instead of a simple string. Seeing a list has been passed, the module would perform the functional equivalent the last example where the module processes the files and submits them in one pass to the K8s cluster.

With this implemented, something like this should be possible with the same performance results:

```
- name: Create the Monitoring Dashboards
  k8s:
    state: "present"
    namespace: "{{ kiali_vars.deployment.namespace }}"
    template: {{ monitoring_dashboard_yamls_to_deploy }}
  when:
    ...
```

Note, here the `monitoring_dashboard_yamls_to_deploy` variable is presumed to contains a list of file paths. This could have been explicitly defined in the task without a variable.

```
- name: Create the Monitoring Dashboards
  k8s:
    state: "present"
[...]

PR #49, opened 2021-04-08
k8s loop flattening for template argument

abikouo:
##### SUMMARY

k8s loop flattening

https://github.com/ansible-collections/community.kubernetes/pull/369
https://github.com/ansible-collections/community.kubernetes/pull/321

Closes: #19
##### ISSUE TYPE

- Feature Pull Request

##### ADDITIONAL INFORMATION

`template` argument can now contains list of item (string or file path)
`continue_on_error` argument is introduced, used with template when multiple resource as defined, defines whether to continue or stop (default behavior) when creation/deletion error occurred.

```paste below

```


changelogs/fragments/49-k8s-loop-flattening-and-continue_on_error.yaml
```diff
@@ -0,0 +1,3 @@
+---
+minor_changes:
+  - k8s - Handle list of definition for option `template`. `continue_on_error` option added (whether to continue on creation/deletion errors) (https://github.com/ansible-collections/kubernetes.core/pull/49).
```
gravesm, line 3: Could you split these two sentences out into separate list items? These are two separate, but related, new features.

plugins/modules/k8s.py
```diff
@@ -123,6 +124,12 @@
       When set to C(yes) leading spaces and tabs are stripped from the start of a line to a block.
       This functionality requires Jinja 2.7 or newer. Default value is false.'
     type: raw
+  continue_on_error:
+    description:
+    - Whether to continue on creation/deletion errors when multiple resources are defined.
+    - This does not include validation step which is controls with C(validate.fail_on_error).
```
gravesm, line 130:
```suggestion
    - This has no effect on the validation step which is controlled by the C(validate.fail_on_error) parameter.
```

gravesm (changes requested):
A few small requests, but this looks good. Nice work.

changelogs/fragments/49-k8s-loop-flattening-and-continue_on_error.yaml
abikouo: @gravesm will do it. thanks

plugins/modules/k8s.py
```diff
@@ -123,6 +124,12 @@
       When set to C(yes) leading spaces and tabs are stripped from the start of a line to a block.
       This functionality requires Jinja 2.7 or newer. Default value is false.'
     type: raw
+  continue_on_error:
```
gravesm, line 127: Just realized we should add a `version_added` field here and set it to `2.0.0`.

gravesm (approved):

gravesm:
@Akasurde @fabianvf can you take a look at this?

plugins/modules/k8s.py
```diff
@@ -123,6 +124,13 @@
...
+    description:
+    - Whether to continue on creation/deletion errors when multiple resources are defined.
+    - This has no effect on the validation step which is controlled by the C(validate.fail_on_error) parameter.
+    type: bool
+    default: False
+    version_added: 2.0.0
```
abikouo, line 133: @goneri

goneri:
Thank you for the patch. @abikouo, do you observe a performance improvement with the patch?

abikouo:
> Thank you for the patch. @abikouo, do you observe a performance improvement with the patch?

@goneri  I did not really take the time to benchmark it, but it seems better that the previous way of using api

plugins/module_utils/common.py
```diff
@@ -616,13 +639,23 @@ def perform_action(self, resource, definition):
                         k8s_obj = resource.delete(**params)
                         result['result'] = k8s_obj.to_dict()
                     except DynamicApiError as exc:
-                        self.fail_json(msg="Failed to delete object: {0}".format(exc.body),
-                                       error=exc.status, status=exc.status, reason=exc.reason)
+                        msg = "Failed to delete object: {0}".format(exc.body)
+                        if continue_on_error:
```
goneri, line 643: I this this pattern coming several times:
```
if continue_on_error:
    blabla
else:
    self.fail_json
```

Can you wrap them in a function?

goneri (changes requested):

plugins/action/k8s_info.py
```diff
@@ -92,22 +92,38 @@ def load_template(self, template, new_module_args, task_vars):
...
+                template_param.update(dict(lstrip_blocks=boolean(template_args.get('lstrip_blocks', False), strict=False)))
             except TypeError as e:
                 raise AnsibleActionFail(to_native(e))
 
-            newline_sequence = template_args.get('newline_sequence', self.DEFAULT_NEWLINE_SEQUENCE)
-            variable_start_string = template_args.get('variable_start_string', None)
-            variable_end_string = template_args.get('variable_end_string', None)
-            block_start_string = template_args.get('block_start_string', None)
-            block_end_string = template_args.get('block_end_string', None)
+            template_param.update(dict(newline_sequence=template_args.get('newline_sequence', self.DEFAULT_NEWLINE_SEQUENCE)))
```
goneri, line 100: Two thing here:
- Can you please use the more Pythonic `{}` instead of `dict()`.
- You can do on single update() if you pass a multi keys `dict. update({"newline_sequence": blabla, "variable_start_string": blabla, "variable_end_string": blabla})`.

goneri (changes requested):

plugins/action/k8s_info.py
```diff
@@ -92,22 +92,38 @@ def load_template(self, template, new_module_args, task_vars):
                     if value is not None and not isinstance(value, string_types):
                         raise AnsibleActionFail("%s is expected to be a string, but got %s instead" % (s_type, type(value)))
             try:
-                trim_blocks = boolean(template_args.get('trim_blocks', True), strict=False)
-                lstrip_blocks = boolean(template_args.get('lstrip_blocks', False), strict=False)
+                template_param.update(dict(trim_blocks=boolean(template_args.get('trim_blocks', True), strict=False)))
```
goneri, line 95: Same thing here. You can use one single update() call and {} instead of dict().

goneri (changes requested):

plugins/action/k8s_info.py
```diff
@@ -92,22 +92,38 @@ def load_template(self, template, new_module_args, task_vars):
...
+        elif isinstance(template, list):
+            for element in template:
+                template_params.append(self.get_template_args(element))
         else:
             raise AnsibleActionFail("Error while reading template file - "
                                     "a string or dict for template expected, but got %s instead" % type(template))
 
         # Option `lstrip_blocks' was added in Jinja2 version 2.7.
-        if lstrip_blocks:
+        if any([tmp['lstrip_blocks'] for tmp in template_params]):
```
goneri, line 126: - I think you can move this block in a new function just to isolate the logic.
- Core has a similar block that also handle Jinja2 <2.7: https://github.com/ansible/ansible/blob/devel/lib/ansible/plugins/action/template.py#L66-L76

goneri (changes requested):

molecule/default/tasks/template.yml
```diff
@@ -148,20 +147,115 @@
...
+
+    - name: validate resource creation succeeded for some and failed for others
+      assert:
+        that:
+          - resource is successful
+          - resource_changed | length > 0
+          - resource_failed | length > 0
+      vars:
+        resource_changed: "{{ resource.result.results | json_query('[?changed==`true`]') }}"
+        resource_failed: "{{ resource.result.results | json_query('[?error]') }}"
```
gravesm, line 243: I would prefer we don't add jmespath as a test dependency just for one assertion. The same should be doable using just the builtin jinja filters. I think this should do the same:
```
assert:
  that:
    - resource.result.results | selectattr('changed') | length == 1
    - resource.result.results | selectattr('error', 'defined') | length == 1
```

gravesm (changes requested):

gravesm (approved):

PR #62, opened 2021-04-15
Add support for tags to molecule tests

gravesm:
##### SUMMARY

Parts of the molecule test suite can now be run selectively using tags.
All the kubernetes related tests are grouped into a tag called `k8s`
while the helm tests are grouped into a tag called `helm`. The k8s tests
are further divided into tags representing the various files in the
tasks folder.

For example, to run only the tests in the waiter.yml file, you can use:
```
  $ molecule converge -- --tags waiter
```

Also, some of k8s tests have not been getting run because they weren't
included anywhere in the test suite. Those are all now included in
converge.yml.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

molecule

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/changelogs/fragments/62-molecule-tags.yaml
+++ b/changelogs/fragments/62-molecule-tags.yaml
@@ -1,0 +1,3 @@
+---
+minor_changes:
+  - add support for using tags when running molecule test suite (https://github.com/ansible-collections/kubernetes.core/pull/62).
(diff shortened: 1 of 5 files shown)


abikouo:
@gravesm really nice to have such a feature!!
would be good to fine tuned also helm tests to add tags for each functionality like on k8s

gravesm:
I wanted to do helm too, but unfortunately the way those tests are structured make it hard to split them apart this way.

goneri (approved):

PR #71, opened 2021-04-20
common: Handle aliases correctly

Akasurde:
##### SUMMARY

Fixed get_api_client API to handle `aliases` that
are passed via inventory configuration or module parameters.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
changelogs/fragments/honor_aliases.yml
molecule/default/tasks/full.yml
plugins/module_utils/common.py

### DIFF
--- a/changelogs/fragments/honor_aliases.yml
+++ b/changelogs/fragments/honor_aliases.yml
@@ -1,0 +1,2 @@
+bugfixes:
+- common - handle ``aliases`` passed from inventory or module params.
--- a/molecule/default/tasks/full.yml
+++ b/molecule/default/tasks/full.yml
@@ -10,6 +10,20 @@
       debug:
         var: output
 
+    - name: Use aliases and check if those values are picked up
+      k8s:
+        name: testing
+        kind: Namespace
+        validate_certs: yes
+        ssl_ca_cert: /dev/null  # invalid CA certificate
+      ignore_errors: yes
+      register: output
+
+    - name: assert that ssl_ca_cert caused a failure (and therefore was correctly translated to ssl_ca_cert)
+      assert:
+        that:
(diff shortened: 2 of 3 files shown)


gravesm:
I don't understand, shouldn't ansible already handle this? If I pass `ssl_ca_cert` and then do `module.params.get('ca_cert')`, won't it already work without this change? If I remove your changes, the tests still pass.

abikouo (changes requested):
> I don't understand, shouldn't ansible already handle this? If I pass `ssl_ca_cert` and then do `module.params.get('ca_cert')`, won't it already work without this change? If I remove your changes, the tests still pass.



> I don't understand, shouldn't ansible already handle this? If I pass `ssl_ca_cert` and then do `module.params.get('ca_cert')`, won't it already work without this change? If I remove your changes, the tests still pass.

I think using module it should work, assuming the core engine will convert argument from aliases to true name
but for inventory plugin, there is no such a conversion, so this has to be done manually.
We should add some ``inventory`` test cases

fabianvf:
note: `AUTH_ARG_MAP` doesn't define aliases for ansible, it's used for mapping the Ansible keys to the proper property names in the kubernetes client. It was added because things like `ssl_ca_cert` were normalized across ansible modules, but that made it so that we couldn't just iterate over the auth_args and set the properties in the kubernetes.Configuration like we were before the normalization. So the `true_name` doesn't need to be exposed to the user, we just need to know what property in the kubernetes.Configuration object the keys that ansible is giving us map to.

Typing this all out, I'm feeling like maybe I should have left this as a comment way back when I added it 😅

gravesm:
@Akasurde What do you think about adding some inventory test cases?

PR #76, opened 2021-04-21
k8s_exec: reuse existing connection

goneri:
Initialize the CoreV1Api object with the existing Kubernetes connection.

### DIFF
--- a/plugins/modules/k8s_exec.py
+++ b/plugins/modules/k8s_exec.py
@@ -144,7 +144,7 @@ def argspec():
 def execute_module(module, k8s_ansible_mixin):
 
     # Load kubernetes.client.Configuration
-    api = core_v1_api.CoreV1Api()
+    api = core_v1_api.CoreV1Api(k8s_ansible_mixin.client.client)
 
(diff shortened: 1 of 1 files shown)


plugins/modules/k8s_exec.py
```diff
@@ -144,7 +144,7 @@ def argspec():
 def execute_module(module, k8s_ansible_mixin):
 
     # Load kubernetes.client.Configuration
-    api = core_v1_api.CoreV1Api()
+    api = core_v1_api.CoreV1Api(k8s_ansible_mixin.client.client)
```
Akasurde, line 147: Could you please add a changelog and test for this change? Thanks,

Akasurde (changes requested):

plugins/modules/k8s_exec.py
goneri: Hi @Akasurde, The change will come when the turbo mode will be merged. Regarding the changelog, this modification is actually a fix of the large pre-turbo mode refactoring, which is still unreleased.

gravesm (approved):

PR #86, opened 2021-04-27
drop python 2 support (breaking changes)

abikouo:
##### SUMMARY
Drop python 2 support for this collection.

### DIFF
--- a/changelogs/fragments/86_drop_python2_support.yaml
+++ b/changelogs/fragments/86_drop_python2_support.yaml
@@ -1,3 +1,3 @@
 ---
-minor_changes:
+breaking_changes:
   - Drop python 2 support.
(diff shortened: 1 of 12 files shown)


changelogs/fragments/86_drop_python2_support.yaml
```diff
@@ -0,0 +1,3 @@
+---
+minor_changes:
```
gravesm, line 2:
```suggestion
breaking_changes:
```

gravesm (changes requested):
Looks like you missed one in the requirements section of plugins/lookup/k8s.py

gravesm:
plugins/lookup/k8s.py still shows a python 2.7 requirement.

gravesm (approved):

PR #89, opened 2021-04-28
Replicate base resource for lists functionality

gravesm:
##### SUMMARY

This replicates specific functionality from the openshift client to more
reliably retrieve the base resource from a resource list.

This addresses part of #34. The specific change to the underlying client being made is here: https://github.com/kubernetes-client/python-base/pull/186/files. The rest of the changes are made to deal with the inheritance chain and hardcoding of the class.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

N/A
##### ADDITIONAL INFORMATION

```paste below

```


tests/sanity/ignore-2.10.txt
```diff
@@ -9,4 +9,8 @@ molecule/default/roles/helm/files/appversionless-chart/templates/configmap.yaml
 molecule/default/roles/helm/files/test-chart-v2/templates/configmap.yaml yamllint!skip
 molecule/default/roles/helm/files/test-chart/templates/configmap.yaml yamllint!skip
 plugins/module_utils/k8sdynamicclient.py import-2.7!skip
-plugins/module_utils/k8sdynamicclient.py import-3.7!skip
```
goneri, line 12: I'm surprised with cannot import with Python 3.7. Do we use some python 3.8+ features?

gravesm: I think this is needed because of the bare imports in the new files. We had to do the same thing previously for the k8sdynamicclient.py.

alinabuzachis (commented):
I would suggest adding some unit tests, what do you think about? https://github.com/openshift/openshift-restclient-python/blob/master/test/unit/test_discoverer.py

gravesm:
@alinabuzachis good suggestion. I had to make a few changes to get the unit tests to work.

plugins/module_utils/client/discovery.py
```diff
@@ -0,0 +1,160 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
```
goneri, line 17: These two lines are for py2. We can safely drop them now.

plugins/module_utils/client/discovery.py
```diff
@@ -0,0 +1,160 @@
...
+        self.__init_cache()
+
+    def __get_default_cache_id(self):
+        user = self.__get_user()
+        if user:
+            cache_id = "{0}-{1}".format(self.client.configuration.host, user)
+        else:
+            cache_id = self.client.configuration.host
+
+        if six.PY3:
```
goneri, line 49: Do we need this if we drop py2 support?

plugins/module_utils/client/discovery.py
```diff
@@ -0,0 +1,160 @@
...
+            return cache_id.encode('utf-8')
+        return cache_id
+
+    def __get_user(self):
+        if hasattr(os, 'getlogin'):
```
goneri, line 54: Why do we need this construction. As far as I can see, `getlogin` is always defined.

goneri: Also, the documentation advice to use `https://docs.python.org/3.6/library/getpass.html#getpass.getuser` which actually can also do the environment variable look up too.

plugins/module_utils/client/discovery.py
gravesm: The sanity tests (for 3.7) fail if they aren't in there. How do I turn those off?

plugins/module_utils/client/discovery.py
```diff
@@ -0,0 +1,160 @@
...
+        try:
+            resources_response = self.client.request('GET', path).resources or []
+        except ServiceUnavailableError:
+            resources_response = []
+
+        resources_raw = list(filter(lambda resource: '/' not in resource['name'], resources_response))
+        subresources_raw = list(filter(lambda resource: '/' in resource['name'], resources_response))
+        for subresource in subresources_raw:
+            resource, name = subresource['name'].split('/')
+            if not subresources.get(resource):
```
goneri, line 107: I would suggest to use a `collections.defaultdict` type https://docs.python.org/3/library/collections.html#collections.defaultdict for `subresources`, and avoid this two lines.

plugins/module_utils/client/discovery.py
gravesm: Nope. I'll remove it. At the time I made this we still had py2 tests being run, but that pr just got merged.

plugins/module_utils/client/resource.py
```diff
@@ -0,0 +1,49 @@
+# Copyright [2017] [Red Hat, Inc.]
```
goneri, line 1: 2021?

tests/unit/module_utils/test_discoverer.py
```diff
@@ -0,0 +1,147 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
```
goneri, line 17: The same, we can drop the lines.

goneri (approved):
Nothing really important, LGTM.

plugins/module_utils/client/discovery.py
goneri: I believe it's these two lines:
https://github.com/ansible-collections/vmware.vmware_rest/blob/main/tests/sanity/ignore-2.11.txt#L17-L18

plugins/module_utils/client/discovery.py
gravesm: This was a copy/paste from the openshift client. From what I can tell it was added to provide support for Windows. I agree it could probably be changed, but I'm hesitant to do so considering we don't do any portability testing and I have no easy way to test any change locally on Windows. Ideally, this code is going to be removed once we can get these or similar changes upstream in the kubernetes client.

goneri: May I suggest to add a comment in the code about that.

alinabuzachis (approved):

plugins/module_utils/k8sdynamicclient.py
```diff
@@ -21,9 +21,13 @@
 
 from ansible_collections.kubernetes.core.plugins.module_utils.apply import apply
 from ansible_collections.kubernetes.core.plugins.module_utils.exceptions import ApplyException
+from ansible_collections.kubernetes.core.plugins.module_utils.client.discovery import LazyDiscoverer
 
 
 class K8SDynamicClient(DynamicClient):
+    def __init__(self, client, cache_file=None, discoverer=None):
+        super().__init__(client, cache_file, discoverer=LazyDiscoverer)
```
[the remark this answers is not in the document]
gravesm: I think you may be looking an older version of this PR. This was removed a few commits ago.

plugins/module_utils/client/resource.py
gravesm: To be honest, no idea. I'm going off what I think @Akasurde said in earlier PRs for this client migration work.

Issue #84, opened 2021-04-26
support "patch only" feature in k8s


jmazzitelli:
I discussed this with @gravesm and @fabianvf - they suggested I submit this enhancement request.

##### SUMMARY
Need a way for `k8s` task to support "patch only". Do not create the resource if it doesn't exist - silently skip that resource (do not raise an error).

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
k8s

##### ADDITIONAL INFORMATION

I have a interesting use-case that `k8s` does not support.

My ansible operator has permissions to get and list all namespaces. My operator also needs to add labels to some existing namespaces (thus, my operator has permission to "patch" namespaces). My operator has NO need (and should not have) permission to create or delete namespaces.

The only way I know of to add or remove labels from a huge list of namespaces most efficiently is to use `state: present` via `k8s` task like this:
```
- name: REMOVE a label from a big list of namespaces by setting label to null but setting state to present
  k8s:
    state: present
    definition: |
      {% for namespace in big_list_of_hundreds_if_not_more_namespaces %}
      {% if namespace in all_known_existing_namespaces_in_the_cluster %}
      ---
      apiVersion: v1
      kind: Namespace
      metadata:
        name: "{{ namespace }}"
        labels:
          my_operator_label: null
      {% endif %}
      {% endfor %}
```

Note that if I set `my_operator_label: "some-non-null-string"` that task ADDS the label to the namespaces. So that is how you add and remove labels via `k8s`.

Here's my problem - I only want this to operate on namespaces that exist - if a namespace does NOT exist, I DO NOT want that namespace created (indeed, my operator can't create it - it doesn't have permission to do so, but even if it did, it would be the wrong thing to do to add it because most likely at this point the user may have deleted the namespace on purpose). But I cannot stop the `k8s` task from trying to create the namespace because `state: present` is telling it to create the namespace if it doesn't exist.

You can see above that I try to work around this - before this task executes, my operator gets a full list of all the namespaces on the cluster - and the if-stmt checks to make sure that namespace exists - if it doesn't, just skip it. But the problem here is that this list can be so long (my QE folks are testing 2,000 namespaces right now) it can take 10 minutes to do this for-loop... and in the timeframe between when the operator queries and obtains the full list of namespaces to the time the `k8s` task gets to a specific namespace to add/remove the label, the user may have deleted that namespace (and as I say, that timeframe could be up to 10 minutes).

[...]

PR #90, opened 2021-04-29
k8s - patch existing resource

abikouo:
##### SUMMARY

add new value ``patched`` for  ``state`` option for ``k8s`` module. patched existing resource with new ``definition``

#84 
##### ISSUE TYPE

- Feature Pull Request
##### COMPONENT NAME

k8s
##### ADDITIONAL INFORMATION

```paste below
kubernetes.core.k8s:
  state: patched
  wait: yes
  definition: |
    ---
    apiVersion: v1
    kind: Namespace
    metadata:
      name: "{{ patch_only_namespace.first }}"
      labels:
        patch: ansible
    ---
    apiVersion: v1
    kind: Namespace
    metadata:
      name: "{{ patch_only_namespace.second }}"
      labels:
        patch: ansible
```


gravesm:
This is probably the cleanest, easiest solution to the immediate problem. I can't think of a better one, but I also have some concerns.

My first concern is that this feels like it kind of goes against how Ansible usually operates. We're using `state: present` which would typically mean creating a thing, but then providing a second parameter to essentially override that. I think this makes the module somewhat harder to use correctly and could lead to more frequent user error.

My second concern is that while this is backwards compatible, which is great, the default behavior of the module is still broken. A patch is designed to modify an existing resource. IMO if the user has supplied a patch our default should be to not create the resource.

tima:
Well reasoned @gravesm. 

To the first concern, I agree while it works it does make the module interface less elegant and more difficult to use. Is there some "prior art" example in the Ansible module space of a state that is "make a change, but don't create it" we could implement here? So I agree. Words matter and I am looking for the right word for this state.

To the second concern, I see what you mean. This is a tough one in that the default behavior is indeed broken so do we fix that behavior and potentially break users who expect it to work that way OR do we let it continue as is and document the inconsistency? I suppose a middle ground is to call it deprecated behavior and emit a warning that they need to switch to what ever we come up with.

tima:
I was browsing modules looking for one that has something that isn't just present and absent. I came across [ansible.windows.win_updates](https://docs.ansible.com/ansible/latest/collections/ansible/windows/win_updates_module.html) that has "installed" as a state. While you don't install a resource, you do defined one. So would `state: defined` be acceptable? Too confusing to separate from absent? 

I haven't had a lot of luck finding alternatives so far otherwise.

gravesm:
A couple other examples with more granular state parameters:
* `file`: absent, directory, file, hard, link, touch
* `apt`: absent, build-dep, latest, present, fixed
* `dnf`: absent, present, installed, removed, latest

I think there's definitely prior art here. If we wanted to maintain backwards compatibility we could perhaps add a new state called `patched` that would only modify existing resources?

Akasurde:
@gravesm I totally agree with you. We can have an additional parameter like `action=patched` with `state=present` so that we align with Ansible Standards. vmware_rest uses this kind of thing where we have `state` different than `absent`, `present`.

gravesm:
@Akasurde could you clarify? Are you suggesting we should add a new state `patched` or are you suggesting we should add a new parameter `action=patched` with `state=present`.

gravesm (approved):

Akasurde:
> @Akasurde could you clarify? Are you suggesting we should add a new state `patched` or are you suggesting we should add a new parameter `action=patched` with `state=present`.

we should add a new parameter `action=patched` with `state=present` if everyone agrees.

abikouo:
> > @Akasurde could you clarify? Are you suggesting we should add a new state `patched` or are you suggesting we should add a new parameter `action=patched` with `state=present`.
> 
> we should add a new parameter `action=patched` with `state=present` if everyone agrees.

we are increasing the complexity of this module, we should rather go for something different, for instance, a new module dealing with all  patching functionalities

Akasurde:
I am OK with new module as well.

jmazzitelli:
`state=patched` seems fine to me. With that set, the context should be clear - if the object doesn't exist, it won't be created nor will that be considered an error condition. It's just "patch an existing resource". That seems clear to me from an end-user (assuming that IS what we are talking about - if it is not, then it is not clear :))

molecule/default/tasks/patched.yml
```diff
@@ -0,0 +1,107 @@
...
+          first: patched-namespace-1
+          second: patched-namespace-2
+
+    - name: Ensure namespace {{ patch_only_namespace.first }} exist
+      kubernetes.core.k8s:
+        definition:
+          apiVersion: v1
+          kind: Namespace
+          metadata:
+            name: "{{ patch_only_namespace.first }}"
```
jmazzitelli, line 14: Could I ask that this test check to ensure that other parts of the resource are not touched (i.e. left as-is). For example, adding a label and annotation and then later testing that that label and annotation still exist after the patch:

```suggestion
            name: "{{ patch_only_namespace.first }}"
            labels:
              existingLabel: "labelValue"
            annotations:
              existingAnnotation: "annotationValue"
```

molecule/default/tasks/patched.yml
```diff
@@ -0,0 +1,107 @@
...
+            labels:
+              ansible: patched
+      register: patch_resource
+
+    - name: assert that patch succeed
+      assert:
+        that:
+          - patch_resource.changed
+          - patch_resource.result.results | selectattr('warning', 'defined') | list | length == 1
+
```
jmazzitelli, line 54: This code might not be exactly right - but hopefully you get the idea. Get the first namespace and make sure the labels and annotations are as expected (you should see the old label and old annotation as well as the new label).

```suggestion

    - name: Ensure namespace {{ patch_only_namespace.first }} was patched correctly
      kubernetes.core.k8s_info:
        kind: namespace
        name: "{{ patch_only_namespace.first }}"
      register: first_namespace

    - name: assert labels are as expected
      assert:
        that:
          - first_namespace.resources[0].metadata.labels.ansible == "patched"
          - first_namespace.resources[0].metadata.labels.existingLabel == "labelValue"
          - first_namespace.resources[0].metadata.annotations.existingAnnotation == "annotationValue"
```

jmazzitelli (commented):
I would suggest adding some tests to make sure the patch does what was expected (check to make sure existing data remains intact (e.g. existing labels and annotations) and see the new label was added).

PR #92, opened 2021-05-03
Remove jsonpatch dependency

gravesm:
##### SUMMARY

The recently discovered bug with json patch merge type was initially
addressed by applying the json patch client-side. This change instead
ensures that the json patch is passed through to the kubernetes api
unmodified. As this addresses the root cause of the bug, the jsonpatch
dependency is no longer necessary.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

k8s
##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/plugins/module_utils/common.py
+++ b/plugins/module_utils/common.py
@@ -106,17 +106,6 @@ except ImportError as e:
     K8S_IMP_ERR = traceback.format_exc()
 
 
-JSON_PATCH_IMP_ERR = None
-try:
-    import jsonpatch
-    HAS_JSON_PATCH = True
-    jsonpatch_import_exception = None
-except ImportError as e:
-    HAS_JSON_PATCH = False
-    jsonpatch_import_exception = e
-    JSON_PATCH_IMP_ERR = traceback.format_exc()
-
-
 def configuration_digest(configuration):
     m = hashlib.sha256()
     for k in AUTH_ARG_MAP:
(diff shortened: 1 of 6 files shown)


plugins/module_utils/common.py
```diff
@@ -537,30 +526,37 @@ def execute_module(self):
...
-
-        for (resource, definition) in flattened_definitions:
-            kind = definition.get('kind', self.kind)
-            api_version = definition.get('apiVersion', self.api_version)
-            definition = self.set_defaults(resource, definition)
-            self.warnings = []
-            if self.params['validate'] is not None:
-                self.warnings = self.validate(definition)
-            result = self.perform_action(resource, definition)
+        if self.params["merge_type"] and self.params["merge_type"][0] == "json":
```
abikouo, line 529: what if ``merge_type`` is defined like this : 
```
- merge
- json
```
according to the documentation 
``If more than one merge_type is given, the merge_types will be tried in order``

abikouo (changes requested):

plugins/module_utils/common.py
gravesm: So the problem here is that `resource_definition` is usually treated as a list of definitions that are processed individually, but a json patch is a list of operations that need to be processed as one. I'm not sure how we support this without a complete rewrite of the module, and even then, I think this is going to be difficult. It's hard for me to think of a good reason why someone would list both json patch and either json merge or strategic merge for `merge_type` as the patch formats are completely different.

I'm open to suggestions for how to address this.

gravesm:
Closing this in favor of a new json patch module.

PR #96, opened 2021-05-05
Replace openshift client with kubernetes client

### LINKED ISSUE #34: Migration to kubernetes-client/python Meta-issue
This meta-issue is part of the execution of the [proposal to establish a separate collection for OpenShift (OKD) Content](#175 ).

The Ansible k8s_* modules and other plugins use the  [openshift-restclient-python library](https://github.com/openshift/openshift-restclient-python) dynamic API client. While this library covers both native K8s functionality in addition to OpenShift specific extensions effectively, the requirement of the library raises questions and concerns with the broader Kubernetes community to its compatibility with all Kubernetes distributions.

This proposed action will make it easier and more effectively to serve these two distinct though related user communities. Switching to the official Kubernetes client and the decoupling of all OpenShift code will assure compatibility for their respective objectives and enable independent decision making on the features, functionality and release cadence.

This meta-issue is tracking the replacement the [openshift-restclient-python library](https://github.com/openshift/openshift-restclient-python) dependency with [the official python client library for Kubernetes](https://github.com/kubernetes-client/python) porting and refactoring code as needed to maintain the same functional behaviors. 

This work is slated to be completed for the v2.0.0 release of this collection.

- [x] Hash generation for configmaps and secrets 
    - Should propose to Kubernetes Python client, this may be of interest to them
- [x] Client-side apply implementation
    - This could partially be solved by adding support for server-side apply, which is available on newer versions of Kubernetes. 
    - It will likely be difficult to upstream this into the Kubernetes Python client, because of the existence of server-side apply
- [x] Advanced diffing logic
    - This is currently part of apply, and we'd likey have to carry this in the collection as this is likely not a feature the Kubernetes Python client maintainers are interested in.
- [x] ~~Support for the `dry_run` query parameter~~ (not currently used)
- [x] Error handling for 503s in the discovery API (https://github.com/kubernetes-client/python-base/pull/187)
- [x] Namespace cache based on user (https://github.com/kubernetes-client/python-base/pull/188)
- [x] Robustly handle *List kind base resource lookups (https://github.com/kubernetes-client/python-base/pull/186)

gravesm:
##### SUMMARY

This commit primarily just removes mentions of openshift from the docs
and updates the requirements. Most of the work to replace the client has
been done through the following commits:

edc48ee5770293af78ac19ac752e7d422b402dae
c214376cac0a54d2be859e37ff7cefa2e4196e7a
48c5170018b7e1869a53588e3b8efee3d93c95aa
2b6a989cf95848fc69c073f9a13f87279c889732

Closes #34.

##### ISSUE TYPE

- Bugfix Pull Request

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/changelogs/fragments/96-replace-openshift-client.yaml
+++ b/changelogs/fragments/96-replace-openshift-client.yaml
@@ -1,0 +1,3 @@
+---
+minor_changes:
+  - replaces the openshift client with the official kubernetes client (https://github.com/ansible-collections/kubernetes.core/issues/34).
--- a/changelogs/fragments/96-replace-openshift-client.yaml
+++ b/changelogs/fragments/96-replace-openshift-client.yaml
@@ -1,3 +1,3 @@
 ---
-minor_changes:
+major_changes:
   - replaces the openshift client with the official kubernetes client (https://github.com/ansible-collections/kubernetes.core/issues/34).
--- a/plugins/lookup/k8s.py
+++ b/plugins/lookup/k8s.py
@@ -125,7 +125,7 @@ DOCUMENTATION = '''
 
     requirements:
       - "python >= 3.6"
-      - "kubernetes >= 11.0.0"
+      - "kubernetes >= 12.0.0"
       - "PyYAML >= 3.11"
 '''
 
--- a/plugins/lookup/k8s.py
+++ b/plugins/lookup/k8s.py
@@ -30,7 +30,7 @@ DOCUMENTATION = '''
       - Fabian von Feilitzsch <@fabianvf>
 
     description:
-      - Uses the OpenShift Python client to fetch a specific object by name, all matching objects within a
+      - Uses the Kubernetes Python client to fetch a specific object by name, all matching objects within a
         namespace, or all matching objects for all namespaces, as well as information about the cluster.
       - Provides access the full range of K8s APIs.
       - Enables authentication via config file, certificates, password or token.
@@ -85,7 +85,7 @@ DOCUMENTATION = '''
       kubeconfig:
         description:
         - Path to an existing Kubernetes config file. If not provided, and no other connection
-          options are provided, the openshift client will attempt to load the default
+          options are provided, the Kubernetes client will attempt to load the default
           configuration file from I(~/.kube/config). Can also be specified via K8S_AUTH_KUBECONFIG environment
           variable.
       context:
@@ -125,13 +125,8 @@ DOCUMENTATION = '''
 
     requirements:
       - "python >= 3.6"
-      - "openshift >= 0.6"
+      - "kubernetes >= 11.0.0"
       - "PyYAML >= 3.11"
-
-    notes:
-      - "The OpenShift Python client wraps the K8s Python client, providing full access to
-        all of the APIS and models available on both platforms. For API version details and
-        additional information visit https://github.com/openshift/openshift-restclient-python"
 '''
 
 EXAMPLES = """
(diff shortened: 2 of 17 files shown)


abikouo (approved):

plugins/lookup/k8s.py
```diff
@@ -125,13 +125,8 @@
 
     requirements:
       - "python >= 3.6"
-      - "openshift >= 0.6"
+      - "kubernetes >= 11.0.0"
```
Akasurde, line 128: @gravesm Do you think we should bump this version to 12.0.0? Since this version contains a major change in the way of setting the config.

Akasurde (changes requested):

changelogs/fragments/96-replace-openshift-client.yaml
```diff
@@ -0,0 +1,3 @@
+---
+minor_changes:
```
Akasurde, line 2:
```suggestion
major_changes:
```

Akasurde: Since we changed the underlying library itself.

plugins/modules/k8s.py
```diff
@@ -52,11 +45,9 @@
     - For example, Custom Resource Definitions typically aren't updatable by the usual strategic merge. You may
       want to use C(merge) if you see "strategic merge patch format is not supported"
     - See U(https://kubernetes.io/docs/tasks/run-application/update-api-object-kubectl-patch/#use-a-json-merge-patch-to-update-a-deployment)
-    - Requires openshift >= 0.6.2
-    - If more than one merge_type is given, the merge_types will be tried in order
-    - If openshift >= 0.6.2, this defaults to C(['strategic-merge', 'merge']), which is ideal for using the same parameters
-      on resource kinds that combine Custom Resources and built-in resources. For openshift < 0.6.2, the default
-      is simply C(strategic-merge).
+    - If more than one merge_type is given, the merge_types will be tried in order. This defaults to
```
Akasurde, line 48:
```suggestion
    - If more than one C(merge_type) is given, the merge_types will be tried in order. This defaults to
```

plugins/modules/k8s_info.py
```diff
@@ -18,7 +18,7 @@
     - "Will Thames (@willthames)"
 
 description:
-  - Use the OpenShift Python client to perform read operations on K8s objects.
+  - Use the Kubernetes  Python client to perform read operations on K8s objects.
```
Akasurde, line 21:
```suggestion
  - Use the Kubernetes Python client to perform read operations on K8s objects.
```

plugins/modules/k8s_service.py
```diff
@@ -33,11 +33,9 @@
     - For example, Custom Resource Definitions typically aren't updatable by the usual strategic merge. You may
       want to use C(merge) if you see "strategic merge patch format is not supported"
     - See U(https://kubernetes.io/docs/tasks/run-application/update-api-object-kubectl-patch/#use-a-json-merge-patch-to-update-a-deployment)
-    - Requires openshift >= 0.6.2
     - If more than one merge_type is given, the merge_types will be tried in order
```
Akasurde, line 36:
```suggestion
    - If more than one C(merge_type) is given, the merge_types will be tried in order
```

Since you are here.

plugins/lookup/k8s.py
gravesm: Didn't we add some logic to handle this? That said, I'm not opposed to bumping our requirement to 12.0.0. I guess now seems like as good a time as any.

alinabuzachis (approved):

PR #99, opened 2021-05-10
Split JSON patch out into a new module

gravesm:
##### SUMMARY

JSON patch in the k8s module has likely never worked. Furthermore,
having resource_definition double as the parameter for patch operations
makes fixing this in k8s difficult. This is due, in part, to the fact
that a single patch consists of a list of operations that should be
processed as a whole, while resource_definition supports a list of
resources that should be processed separately. The merge_type parameter
allows more than one merge strategy to be specified, but there is no
case where a JSON patch could work for JSON merge or strategic merge,
nor the other way around.

This commit moves JSON patch functionality out into a separate module.
Check mode support is provided through the third party jsonpatch
library.

##### ISSUE TYPE

- New Module Pull Request

##### COMPONENT NAME

k8s_json_patch

##### ADDITIONAL INFORMATION

```paste below

```


changelogs/fragments/99-json-patch-module.yaml
```diff
@@ -0,0 +1,3 @@
+---
+minor_changes:
+  - k8s_json_patch - split JSON patch functionality out into a separate module (https://github.com/ansible-collections/kubernetes.core/pull/99).
```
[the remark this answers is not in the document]
gravesm: I have gone ahead and just deprecated `merge_type=json`. While it never worked, it usually failed silently.

jillr (approved):

plugins/modules/k8s_json_patch.py
```diff
@@ -0,0 +1,279 @@
...
+      - op: add
+        path: /metadata/labels/app
+        value: myapp
+      - op: replace
+        patch: /spec/containers/0/image
+        value: nginx
+'''
+
+RETURN = r'''
+result:
```
Akasurde, line 86: `result` is a reserved keyword. I would suggest changing it to something meaningful like `patched_output` or `json_output`.

Akasurde (changes requested):
Rest LGTM. Feel free to merge after resolving the comments.

abikouo:
I am thinking about having a ``k8s_patch`` module instead dealing with all related patch stuffs like the ``kubectl patch`` command. However for the definition used for ``strategic merge`` can also be used to create an object for now, but we should remove this feature to have a module ``k8s_patch`` patching only existing objects and ``k8s`` used to create them.

molecule/default/tasks/json_patch.yml
```diff
@@ -0,0 +1,170 @@
...
+            containers:
+              - image: busybox:musl
+                name: busybox
+                command:
+                  - sh
+                  - -c
+                  - while true; do echo $(date); sleep 10; done
+        wait: yes
+
+    - name: Add a label and replace the image in checkmode
```
fabianvf, line 32: Could we add a case that tests the `remove` action and verifies it's idempotent?

fabianvf (commented):
re: https://github.com/ansible-collections/kubernetes.core/pull/99#discussion_r629846551

One thing to note is that the other modules do return `result` [0][1]. If `result` isn't appropriate we should probably change the return value in the other modules as well for 3.0, and it would be nice if it matched what was output here (so probably not `patched_output`). As a user I think it would be nicer if there was a consistent return value key I could check, rather than having to check/remember what the output key for the same type of information is for each module separately. Perhaps `k8s_object` or something like that would make sense.

0: https://github.com/ansible-collections/kubernetes.core/blob/main/docs/kubernetes.core.k8s_module.rst#return-values
1: https://github.com/ansible-collections/kubernetes.core/blob/main/docs/kubernetes.core.k8s_scale_module.rst#return-values

molecule/default/tasks/json_patch.yml
gravesm: Hmm, I'm not sure if we can test idempotency for remove ops. The path won't exist the second time around and kubernetes will return an error.

gravesm:
@fabianvf false alarm. `result` is not a reserved keyword. Reverting back to `result`.

Akasurde:
LGTM

PR #102, opened 2021-05-11
Account for updated pods when waiting on DaemonSet

gravesm:
##### SUMMARY

We have been seeing infrequent CI failures recently during the DaemonSet tests in the waiter test suite. The problem can be reliably reproduced by using a container that ignores SIGTERM and forces the replacement cycle to run out the grace period for termination. The logic we're using to determine when a DaemonSet is ready checks whether the number of nodes that should be running the daemon pod is the same as the number nodes currently running a daemon pod. In this case, however, the pod is the old pod that just hasn't been replaced, yet. This change makes sure the number of nodes that should be running the daemon pod is the same as the number of nodes currently running _the updated daemon pod_.

From what I can tell, this follows the same logic used by kubectl when waiting on a rollout:
https://github.com/kubernetes/kubectl/blob/ac49920c0ccb0dd0899d5300fc43713ee2dfcdc9/pkg/polymorphichelpers/rollout_status.go#L107-L115.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

k8s

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/changelogs/fragments/102-wait-updated-daemonset-pods.yaml
+++ b/changelogs/fragments/102-wait-updated-daemonset-pods.yaml
@@ -1,0 +1,3 @@
+---
+minor_changes:
+  - k8s - wait for all pods to update when rolling out daemonset changes (https://github.com/ansible-collections/kubernetes.core/pull/102).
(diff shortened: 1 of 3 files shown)


Akasurde (approved):
LGTM

abikouo (approved):
LGTM

PR #105, opened 2021-05-18
k8s ability to wait on arbitrary property

### LINKED ISSUE #66: Ability to wait on arbitrary status value
##### SUMMARY

Not all k8s resources implement the conditions in their status. However it may still be necessary to wait on them being ready. Since operations in K8s are asynchronous, an item later depending on this may fail.

A good example is the kind Service. This does not implement conditions as shown in https://github.com/kubernetes/kubernetes/issues/80828. My ansible may need to wait for this service to have acquired an External IP address before using in a further task - perhaps feeding that IP into a network infrastructure rule (firewall, DNS or other). 

A similar ticket exists for the Kubectl command - https://github.com/kubernetes/kubernetes/issues/83094 - waiting for an arbitrary json path.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

module k8s

##### ADDITIONAL INFORMATION

```yaml
- name: Deploy the dashboard service (lb)
  community.kubernetes.k8s:
    <<: *k8s_auth
    template: dash-service.yaml
    wait: yes
    wait_for: .status.loadBalancer.ingress[*].ip
```

Something like that could be waiting for an ingress IP to appear. Perhaps inspiration could come from https://github.com/kubernetes/kubernetes/issues/83094#issuecomment-715975514.

abikouo:
##### SUMMARY

improve wait for ``k8s`` module.

Closes #66 
##### ISSUE TYPE

- Feature Pull Request
##### COMPONENT NAME

k8s


plugins/module_utils/common.py
```diff
@@ -96,6 +98,13 @@
     k8s_import_exception = e
     K8S_IMP_ERR = traceback.format_exc()
 
+try:
+    import jsonpath_rw
```
fabianvf, line 102: I guess there's probably too much to the jsonpath syntax to not require an additional library, are there other Ansible modules that use this syntax and if so is this the library they use as well?

plugins/doc_fragments/k8s_wait_options.py
```diff
@@ -64,4 +64,9 @@ class ModuleDocFragment(object):
         - The possible reasons in a condition are specific to each resource type in Kubernetes.
         - See the API documentation of the status field for a given resource to see possible choices.
     type: dict
+  wait_for:
```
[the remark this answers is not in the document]
fabianvf: I guess `wait_property` or something but I kind of dislike that more I think

plugins/module_utils/common.py
```diff
@@ -432,19 +441,49 @@ def _custom_condition(resource):
...
+                if condition:
+                    # add waiter on custom condition
+                    predicates.append(_custom_condition)
+                if wait_for:
+                    # json path predicate
+                    predicates.append(_wait_for_property)
         else:
-            predicate = _resource_absent
-        return self._wait_for(resource, definition['metadata']['name'], definition['metadata'].get('namespace'), predicate, sleep, timeout, state)
+            predicates.append(_resource_absent)
```
fabianvf, line 485: I think we'd want to make sure that resource_absent is the only predicate in this case, the resource being waited on should be deleted so any other predicate in addition to that one will probably just make it wait until it times out

plugins/doc_fragments/k8s_wait_options.py
```diff
@@ -64,4 +64,9 @@ class ModuleDocFragment(object):
...
+    description:
+    - Specifies a property on the resource to wait for.
+    - Ignored if C(wait) is not set or is set to I(False).
```
fabianvf, line 70: Some more detailed syntax examples or links to syntax documentation would be helpful think

fabianvf (changes requested):

plugins/doc_fragments/k8s_wait_options.py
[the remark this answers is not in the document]
abikouo: Thanks for the feedback I totally agree with you.

plugins/module_utils/common.py
abikouo: That's was also my concern, I don't really like adding a new library I will try to find if there is such a library already implemented in another plugin

plugins/modules/k8s.py
```diff
@@ -236,6 +236,13 @@
       type: Progressing
       status: Unknown
       reason: DeploymentPaused
+
+# Wait for this service to have acquired an External IP
+- name: Deploy the dashboard service (lb)
+  kubernetes.core.k8s:
+    template: dash-service.yaml
+    wait: yes
+    wait_for: .status.loadBalancer.ingress[*].ip
```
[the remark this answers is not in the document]
abikouo: It was supporting only the attribute present but I will try to add complex checks, so the syntax would be similar to this 
```
wait_property:
    attribute: status.phase
    value: Completed
```
When there is no value specified we do check only if the ``attribute`` is ``present``

plugins/module_utils/common.py
gravesm: I was going to suggest jmespath, too. There is a jmespath filter that used to be in core, but I think recently got moved to community.general. I believe there was some concern about the maintenance of this library, but I'm not sure. jmespath is also a dependency of boto3, so it's at least already a transitive dependency for the aws collection.

plugins/modules/k8s.py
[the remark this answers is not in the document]
gravesm: If we are going to be supporting the ability to wait on arbitrary properties/values, is there any value in continuing to support the `wait_condition` parameter?

plugins/doc_fragments/k8s_wait_options.py
```diff
@@ -64,4 +64,9 @@ class ModuleDocFragment(object):
...
+    type: str
```
Akasurde, line 71:
```suggestion
    type: str
    version_added: '2.0.0'
```

plugins/module_utils/common.py
```diff
@@ -96,6 +98,13 @@
     k8s_import_exception = e
     K8S_IMP_ERR = traceback.format_exc()
 
+try:
+    import jsonpath_rw
+    HAS_JSONPATH_RW = True
+    jsonpath_import_exception = None
+except ImportError as e:
+    HAS_JSONPATH_RW = False
+    jsonpath_import_exception = e
```
Akasurde, line 107:
```suggestion
    JSONPATH_IMP_ERR = e
```

plugins/module_utils/common.py
```diff
@@ -96,6 +98,13 @@
     k8s_import_exception = e
     K8S_IMP_ERR = traceback.format_exc()
 
+try:
+    import jsonpath_rw
+    HAS_JSONPATH_RW = True
+    jsonpath_import_exception = None
```
Akasurde, line 104:
```suggestion
    JSONPATH_IMP_ERR = None
```

plugins/module_utils/common.py
```diff
@@ -432,19 +441,49 @@ def _custom_condition(resource):
         def _resource_absent(resource):
             return not resource
 
+        # wait_for requires jsonpath-rw library
+        jsonpath_expr = None
+        if wait_for is not None:
+            if not HAS_JSONPATH_RW:
+                if hasattr(self, 'fail_json'):
+                    self.fail_json(msg=missing_required_lib('jsonpath_rw'), error=to_native(jsonpath_import_exception))
```
Akasurde, line 449:
```suggestion
                    self.fail_json(msg=missing_required_lib('jsonpath_rw'), error=to_native(JSONPATH_IMP_ERR))
```

plugins/module_utils/common.py
```diff
@@ -241,7 +250,7 @@ def find_resource(self, kind, api_version, fail=False):
                 self.fail(msg='Failed to find exact match for {0}.{1} by [kind, name, singularName, shortNames]'.format(api_version, kind))
 
     def kubernetes_facts(self, kind, api_version, name=None, namespace=None, label_selectors=None, field_selectors=None,
-                         wait=False, wait_sleep=5, wait_timeout=120, state='present', condition=None):
+                         wait=False, wait_sleep=5, wait_timeout=120, state='present', condition=None, wait_for=None):
```
Akasurde, line 253: Add an example about this in the example section.

plugins/filter/k8s.py
```diff
@@ -23,7 +23,7 @@ def k8s_config_resource_name(resource):
 # ---- Ansible filters ----
 class FilterModule(object):
     """
-
+    Kubernetes filter Module
```
Akasurde, line 26: This is not required. Since the bug around this is resolved.

abikouo:
Expecting the following to be merged https://github.com/ansible-collections/cloud.common/pull/56 for the molecule test to be ok

abikouo:
recheck

molecule/default/tasks/waiter.yml
```diff
@@ -364,6 +364,93 @@
...
+        wait_property:
+          property: status.containerStatuses[*].state.running
+      ignore_errors: true
+      register: crash_pod
+
+    - name: assert that task failed
+      assert:
+        that:
+          - crash_pod is failed
+          - crash_pod.changed
```
Akasurde, line 391:
```suggestion
          - crash_pod is changed
```

Akasurde (approved):
LGTM.

gravesm (commented):
I'm not sure how I feel about this one. This is a pretty limited subset of JMESPath and I'm worried if we release this we are going to continually be fielding requests to support a missing syntax feature. This feels like the first step down the road of creating our own implementation of JMESPath. If we decide to move forward with this it is going to at least need more documentation about what specific syntax features are supported.

abikouo:
> I'm not sure how I feel about this one. This is a pretty limited subset of JMESPath and I'm worried if we release this we are going to continually be fielding requests to support a missing syntax feature. This feels like the first step down the road of creating our own implementation of JMESPath. If we decide to move forward with this it is going to at least need more documentation about what specific syntax features are supported.

With I am submitting another solution on how to deal with json path, we need to decide whether use an existing library and add it as dependency or create our own and maintain it

Akasurde:
I feel the current solution is apt for now. Once we get feedback from the user we can think about expanding this feature or using the third-party library.

plugins/module_utils/jsonpath_extractor.py
```diff
@@ -0,0 +1,142 @@
...
+    return search_json_item(jsonpath_expr, data)
+
+
+def validate_with_jsonpath(module, data, expr, value=None):
+    def _raise_or_fail(err, **kwargs):
+        if module and hasattr(module, "fail_json"):
+            module.fail_json(error=to_native(err), **kwargs)
+        raise err
+
+    def _match_value(buf, v):
```
gravesm, line 121: This seems broken when trying to wait on a property that is an integer. For example, I can't get the following to work with a Deployment:
```
wait: yes
wait_property:
  property: status.availableReplicas
  value: 1
```

plugins/module_utils/jsonpath_extractor.py
```diff
@@ -0,0 +1,142 @@
...
+        if isinstance(buf, list):
+            # convert all values from bool to str and lowercase them
+            return all([str(i).lower() == v.lower() for i in buf])
+        elif isinstance(buf, str):
+            return v.lower() == content.lower()
```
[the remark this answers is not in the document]
abikouo: it should not be a breaking change, moving from this library to JMESPath will impact only the way data are returned from JSON document (content).

plugins/module_utils/jsonpath_extractor.py
abikouo: updated

plugins/module_utils/jsonpath_extractor.py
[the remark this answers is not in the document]
gravesm: Yeah, you're right. Was thinking properties, not values.

PR #106, opened 2021-05-18
Remove empty docstring in filter plugin

gravesm:
##### SUMMARY

A recent change somewhere upstream broke the sanity tests. This
partially reverts 0eb1559cc7347398a66bcb6586e37d594d094106 to remove the
empty docstring on the filter plugin. The reason the empty docstring was
added in the first place was to fix a problem with the doc generation
script. However, a recent change to the collection_prep project fixed
that issue so we no longer need the empty docstring.

##### ISSUE TYPE

- Bugfix Pull Request

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/plugins/filter/k8s.py
+++ b/plugins/filter/k8s.py
@@ -22,9 +22,6 @@ def k8s_config_resource_name(resource):
 
 # ---- Ansible filters ----
 class FilterModule(object):
-    """
-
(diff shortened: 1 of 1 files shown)


Akasurde:
@gravesm Thanks for fixing this.

PR #117, opened 2021-06-01
lookup: return list even if single item found

Akasurde:
##### SUMMARY

Always return list from k8s lookup plugin

Fixes: #9

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
changelogs/fragments/9_lookup_k8s.yml
molecule/default/converge.yml
molecule/default/tasks/lookup_k8s.yml
plugins/lookup/k8s.py

### DIFF
--- a/molecule/default/tasks/lookup_k8s.yml
+++ b/molecule/default/tasks/lookup_k8s.yml
@@ -1,0 +1,54 @@
+---
+- block:
+    # https://github.com/ansible-collections/kubernetes.core/issues/9
+    - name: Create a namespace with label
+      kubernetes.core.k8s:
+        definition:
+          apiVersion: v1
+          kind: Namespace
+          metadata:
+            name: "app-development-one"
+            labels:
+              namespace_label: "app_development"
+
+    - set_fact:
+        namespace_info: "{{ lookup('kubernetes.core.k8s', kind='Namespace', label_selector='namespace_label=app_development') }}"
+
+    - name: Check if the returned value is list with a single element
+      assert:
+        that:
+          - namespace_info is iterable
+          - not namespace_info is string
+          - not namespace_info is mapping
+          - namespace_info | length == 1
+
+    - name: Create another namespace with label
+      kubernetes.core.k8s:
+        definition:
+          apiVersion: v1
+          kind: Namespace
+          metadata:
+            name: "app-development-two"
+            labels:
+              namespace_label: "app_development"
+
+    - set_fact:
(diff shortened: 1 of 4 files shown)


molecule/default/tasks/lookup_k8s.yml
```diff
@@ -0,0 +1,30 @@
...
+          - not namespace_info is string
+          - not namespace_info is mapping
+          - namespace_info | length == 1
+
+  always:
+    - name: Ensure that namespace is removed
+      k8s:
+        kind: Namespace
+        name: app_development_one
+        state: absent
```
abikouo, line 30:
```suggestion
    - name: create another namespace with label
      kubernetes.core.k8s:
        definition:
          apiVersion: v1
          kind: Namespace
          metadata:
            name: "app_development_two"
            labels:
              namespace_label: "app_development"

    - set_fact:
        namespace_info: "{{ lookup('kubernetes.core.k8s', kind='Namespace', label_selector='namespace_label=app_development') }}"

    - name: Check if the returned value is list with a two elements
      assert:
        that:
          - namespace_info is iterable
          - not namespace_info is string
          - not namespace_info is mapping
          - namespace_info | length == 2
  always:
    - name: Ensure that namespace is removed
      k8s:
        kind: Namespace
        name: '{{ item }}'
        state: absent
    with_items:
    - app_development_one
    - app_development_two
```

plugins/lookup/k8s.py
```diff
@@ -277,7 +264,11 @@ def run(self, terms, variables=None, **kwargs):
         if self.name:
             return [k8s_obj.to_dict()]
 
-        return k8s_obj.to_dict().get('items')
+        items = k8s_obj.to_dict().get('items')
+
+        if len(items) == 1:
```
abikouo, line 269: checking the ``len`` means that the ``items`` is already a list, otherwise it will check the size of a ``dict`` or ``string`` ? and if it is already a list we can just return it

molecule/default/tasks/lookup_k8s.yml
```diff
@@ -0,0 +1,30 @@
+---
+- block:
+    # https://github.com/ansible-collections/kubernetes.core/issues/9
+    - name: Create a namespace with label
+      kubernetes.core.k8s:
+        definition:
+          apiVersion: v1
+          kind: Namespace
+          metadata:
+            name: "app_development_one"
```
abikouo, line 10:
```suggestion
            name: "app-development-one"
```

plugins/lookup/k8s.py
Akasurde: Good point.

gravesm (approved):

PR #119, opened 2021-06-02
Support multiple resources with same GVK

gravesm:
##### SUMMARY

In replicating the openshift client functionality we forgot to bring
over a fix for multiple resources with the same group version and kind.
This is specifically a problem for openshift templates.

##### ISSUE TYPE

- Bugfix Pull Request

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/plugins/module_utils/client/discovery.py
+++ b/plugins/module_utils/client/discovery.py
@@ -22,7 +22,8 @@ import tempfile
 import kubernetes.dynamic
 import kubernetes.dynamic.discovery
 from kubernetes import __version__
-from kubernetes.dynamic.exceptions import ServiceUnavailableError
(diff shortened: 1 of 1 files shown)


plugins/module_utils/client/discovery.py
```diff
@@ -22,7 +22,8 @@
 import kubernetes.dynamic
 import kubernetes.dynamic.discovery
 from kubernetes import __version__
-from kubernetes.dynamic.exceptions import ServiceUnavailableError
+from kubernetes.dynamic.exceptions import (ResourceNotFoundError, ResourceNotUniqueError,
```
Akasurde, line 25: Add a changelog entry and Rest LGTM

Akasurde (approved):

PR #127, opened 2021-06-10
k8s_cp - a new module for copying files to/from a Pod

abikouo:
##### SUMMARY

``k8s_cp`` is a new module to copy file/directories to/from a Pod

#10 
##### ISSUE TYPE

- New Module Pull Request

##### COMPONENT NAME

k8s_cp
##### ADDITIONAL INFORMATION

```paste below
- name: copy server configuration into pod
  k8s_cp:
     pod: some-server-pod
     namespace: some-namespace
     local_path: files/configuration.yaml
     remote_path: /root/nginx/server.yaml
```


plugins/modules/k8s_cp.py
```diff
@@ -0,0 +1,385 @@
...
+                stderr.append(resp.read_stderr())
+        err = resp.read_channel(3)
+        err = yaml.safe_load(err)
+        resp.close()
+        if err['status'] == 'Success':
+            return True
+        return False
+
+    def copy_local_file_to_remote(self, local_file, remote_file):
+        tar_command = ['tar', 'xvf', '-', '-C', '/']
```
Akasurde, line 233:
```suggestion
        from ansible.module_utils.common.process import get_bin_path
        try:
            tar_bin = get_bin_path('tar')
        except ValueError:
        	self.module.fail_json("Unable to find the 'tar' binary in the path.")
        tar_command = [tar_bin, 'xvf', '-', '-C', '/']
```

plugins/modules/k8s_cp.py
```diff
@@ -0,0 +1,385 @@
...
+                if commands:
+                    cmd = commands.pop(0)
+                    response.write_stdin(cmd.decode())
+                else:
+                    break
+            response.close()
+            if stderr:
+                self.module.fail_json(msg="copy operation failed: {0}".format(stderr))
+            else:
+                self.module.exit_json(changed=True, result="content successfully copied into file {0} on remote pod".format(self.remote_path))
```
Akasurde, line 264:
```suggestion
            if stderr:
                self.module.fail_json(msg="copy operation failed: {0}".format(stderr))
            self.module.exit_json(changed=True, result="content successfully copied into file {0} on remote pod".format(self.remote_path))
```

plugins/modules/k8s_cp.py
```diff
@@ -0,0 +1,385 @@
...
+        is_remote_path_file = self.test_remote_path(as_dir=False)
+
+        if not is_remote_path_dir and not is_remote_path_file:
+            self.module.fail_json(msg="{0} does not exist in remote pod filesystem".format(self.remote_path))
+
+        is_local_path_dir = os.path.isdir(self.local_path)
+        if is_remote_path_dir and not is_local_path_dir:
+            self.module.fail_json(msg="{0} does not exist in local filesystem".format(self.local_path))
+
+        exec_command = ['tar', 'cf', '-', '-C', os.path.dirname(self.remote_path), os.path.basename(self.remote_path)]
```
Akasurde, line 300: get_bin_path as per above.

plugins/modules/k8s_cp.py
```diff
@@ -0,0 +1,385 @@
...
+    argument_spec = copy.deepcopy(AUTH_ARG_SPEC)
+    argument_spec['namespace'] = {'type': 'str', 'required': True}
+    argument_spec['pod'] = {'type': 'str', 'required': True}
+    argument_spec['container'] = {}
+    argument_spec['remote_path'] = {'type': 'path', 'required': True}
+    argument_spec['local_path'] = {'type': 'path'}
+    argument_spec['content'] = {'type': 'str'}
+    argument_spec['state'] = {'type': 'str', 'default': 'to_pod', 'choices': ['to_pod', 'from_pod']}
+
+    module = AnsibleModule(argument_spec=argument_spec,
```
Akasurde, line 375: Could you please generate docs using `collection_prep_add_docs` command?

Akasurde (changes requested):

plugins/modules/k8s_cp.py
abikouo: This is the command which will be executed on the kubernetes stream, so the tar executable is the one from the pod, so retrieving the bin path from the local host is not what we intend.

plugins/modules/k8s_cp.py
abikouo: same as above

gravesm:
I think this should support transfer of binary data, which it currently does not.

abikouo:
> I think this should support transfer of binary data, which it currently does not.

updated.

molecule/default/roles/k8scopy/files/data/ansible/collection.txt
```diff
@@ -0,0 +1 @@
+kubernets.core
```
Akasurde, line 1:
```suggestion
kubernetes.core
```
Not a blocker

molecule/default/roles/k8scopy/tasks/test_copy_errors.yml
```diff
@@ -0,0 +1,69 @@
+---
+# copy non-existent local file shoud fail
```
Akasurde, line 2:
```suggestion
# copy non-existent local file should fail
```

molecule/default/roles/k8scopy/tasks/test_copy_errors.yml
```diff
@@ -0,0 +1,69 @@
...
+  ignore_errors: true
+  register: copy_non_existent
+
+- name: check that error message is as expected
+  assert:
+    that:
+      - copy_non_existent is failed
+      - copy_non_existent.msg == "this_file_does_not_exist does not exist in local filesystem"
+
+# copy non-existent pod file shoud fail
```
Akasurde, line 19:
```suggestion
# copy non-existent pod file should fail
```

molecule/default/roles/k8scopy/tasks/test_copy_errors.yml
```diff
@@ -0,0 +1,69 @@
...
+  register: copy_non_existent
+
+- name: check that error message is as expected
+  assert:
+    that:
+      - copy_non_existent is failed
+      - copy_non_existent.msg == "/this_file_does_not_exist does not exist in remote pod filesystem"
+
+# copy file into multiple container pod without specifying the container should fail
+- name: copy file into mutiple container pod
```
Akasurde, line 37:
```suggestion
- name: copy file into multiple container pod
```

molecule/default/roles/k8scopy/tasks/test_copy_errors.yml
```diff
@@ -0,0 +1,69 @@
...
+  register: copy_multi_container
+
+- name: check that error message is as expected
+  assert:
+    that:
+      - copy_multi_container is failed
+      - copy_multi_container.msg == "Pod contains more than 1 container, option 'container' should be set"
+
+# copy using non-existent container from pod should failed
+- name: copy file into mutiple container pod
```
Akasurde, line 54:
```suggestion
- name: copy file into multiple container pod
```

plugins/modules/k8s_cp.py
```diff
@@ -0,0 +1,494 @@
...
+__metaclass__ = type
+
+
+DOCUMENTATION = r'''
+
+module: k8s_cp
+
+short_description: Copy files and directories to and from pod.
+
+version_added: "2.1.0"
```
Akasurde, line 16:
```suggestion
version_added: "2.1.2"
```

plugins/module_utils/common.py
```diff
@@ -214,7 +214,7 @@ def __init__(self, module, *args, **kwargs):
                              error=to_native(k8s_import_exception))
         self.kubernetes_version = kubernetes.__version__
 
-        if not HAS_YAML:
+        if kwargs.get('pyyaml_not_required', True) and not HAS_YAML:
```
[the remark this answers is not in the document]
abikouo: this is mandatory for modules using ``common.py`` which do not need ``PyYAML`` library, as this is by default mandatory for all these modules.

abikouo:
> A few things here. It seems a little weird that we are using tar in one direction and cat in another. I think it should probably be the same in both directions. I also think cat is the better way to go. We can't assume a GNU tar on the target container. Alpine, for example, packages its own stripped down version of tar.
> 
> We should also use streaming for all the IO. I would imagine one use case for this module will be to grab log files from a container and those could potentially be large.

We can use ``cat`` in both directory, I need to test if this is working with zip and archive
Regarding the streaming of the input and output I think this should be achieve using ``k8s_exec`` module.

plugins/module_utils/common.py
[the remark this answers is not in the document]
gravesm: The naming of this is very confusing. The way this reads is "if pyyaml is not required, and we don't have pyyaml, then fail." If pyyaml is not required then it shouldn't matter whether or not it's installed.

gravesm:
If we can only get tar working for now, that's fine. If this is the case we need to document that the tar binary is required on the container.

abikouo:
it is working with tar, the tar binary requirement has been added

plugins/module_utils/common.py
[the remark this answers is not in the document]
gravesm: This is still really confusing. Can you either rename this to `pyyaml_required` or find some other way to do this? It leads to what you have down below where you pass `pyyaml_not_required=False` which is the exact opposite of what you want.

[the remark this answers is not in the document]
abikouo: In fact I did it but forgot to push the commit

gravesm (changes requested):
I'm getting a broken pipe error when I try to copy a file to a pod that's larger than around 40MB. My guess is if you chunk the data it will probably fix it.

molecule/default/roles/k8scopy/tasks/test_copy_large_file.yml
```diff
@@ -0,0 +1,120 @@
...
+  - name: Compare files
+    kubectl_file_compare:
+      namespace: '{{ copy_namespace }}'
+      pod: '{{ pod_with_one_container.name }}'
+      remote_path: /large_text_file.txt
+      local_path: "{{ test_directory }}/large_text_file_from_pod.txt"
+      kubectl_path: "{{ kubectl_path }}"
+
+  # Copy large executable file from/to local filesystem to Pod
+  - name: copy large file into remote Pod
```
gravesm, line 66: kind is not a particularly large binary file. I'd suggest instead just creating a random file with something like:
```command: dd if=/dev/urandom of={{ local_path }} bs=1M count=100``` or similar.

Rather than testing the output of the executable we should just be checking the sha1sum of the file. It's possible an executable could still function well enough even if some of it got corrupted in the transfer, especially if we're just checking the version.

**Review by gravesm — CHANGES_REQUESTED** [review]

File: molecule/default/roles/k8scopy/tasks/test_copy_large_file.yml (hunk 12 above)
Comment (abikouo) [reply molecule/default/roles/k8scopy/tasks/test_copy_large_file.yml:66]: I have created a test module to compare the content, I will keep it for executable to compare both content and execution output

**Review by gravesm — CHANGES_REQUESTED** [review]

File: plugins/action/k8s_info.py [hunk 13]
```diff
@@ -238,6 +253,16 @@ def run(self, tmp=None, task_vars=None):
         if template:
             self.load_template(template, new_module_args, task_vars)
 
+        local_path = self._task.args.get('local_path')
+        state = self._task.args.get('state', None)
+        with open('/tmp/k8s_action.txt', 'w') as f:
+            f.write("args {0}".format(self._task.args))
```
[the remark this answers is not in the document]
abikouo: yes, thanks

gravesm (changes requested):
Could you fix the examples in the module? They are all using `k8s` and need to use `k8s_cp`.

abikouo:
> Could you fix the examples in the module? They are all using `k8s` and need to use `k8s_cp`.

Done

gravesm (approved):
Thanks!

itaru2622:
to maintainer:
Could you please release this feature (current main branch) as new TAG as soon as possible?

PR #132, opened 2021-06-15
Fix molecule test ``gc.yml``

abikouo:
##### SUMMARY

Molecule are falling randomly when running tasks ``gc.yml``
Depending on the system the pod is not already running when we try to validate the state.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

Molecule tests

### DIFF
--- a/molecule/default/tasks/gc.yml
+++ b/molecule/default/tasks/gc.yml
@@ -46,6 +46,9 @@
         wait: yes
         wait_timeout: 100
       register: job
+      until: job.resources[0].status.phase == "Running"
+      retries: 5
+      delay: 10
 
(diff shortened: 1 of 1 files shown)


molecule/default/tasks/gc.yml
```diff
@@ -46,6 +46,9 @@
         wait: yes
         wait_timeout: 100
       register: job
+      until: job.resources[0].status.phase == "Running"
```
gravesm, line 49: There are a couple other places in this file that we should probably do the same thing for.

gravesm (approved):

PR #135, opened 2021-06-15
Fix apply on Ansible 2.9

gravesm:
##### SUMMARY

For some reason the apply function can't be correctly imported in
Ansible 2.9. This just renames it to get it to import. I've also added
molecule testing on multiple Ansible versions.

##### ISSUE TYPE

- Bugfix Pull Request

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -75,7 +75,7 @@ jobs:
     strategy:
       matrix:
         python_version: ['3.7']
-        ansible_version: ['==2.9.*', '']
+        ansible_version: ['==2.9.*', '==2.10.*', '']
     steps:
       - name: Check out code
         uses: actions/checkout@v2


plugins/module_utils/k8sdynamicclient.py
```diff
@@ -19,7 +19,7 @@
 
 from kubernetes.dynamic import DynamicClient
 
-from ansible_collections.kubernetes.core.plugins.module_utils.apply import apply
+from ansible_collections.kubernetes.core.plugins.module_utils.apply import k8s_apply
```
abikouo, line 22: ``import apply as k8s_apply`` why not ?

gravesm: This is broken as well. The only way I can get this to work is to just rename the function.

.github/workflows/ci.yml
```diff
@@ -75,6 +75,7 @@ jobs:
     strategy:
       matrix:
         python_version: ['3.7']
+        ansible_version: ['==2.9.*', '']
```
abikouo, line 78:
```suggestion
        ansible_version: ['==2.9.*', '==2.10.*', '==2.11.*', '']
```

plugins/action/k8s_service.py
```diff
@@ -0,0 +1 @@
+k8s_info.py
```
Akasurde, line 1: I didn't realize runtime.yml does not work on 2.9. @gravesm Thanks for catching this.

.github/workflows/ci.yml
gravesm: This won't work because there is no 2.11 release of ansible.

abikouo: so at least 2.10 in order to validate on different version

abikouo (approved):

goneri (approved):

PR #141, opened 2021-06-21
k8s_drain new module

abikouo:
##### SUMMARY

new module to drain, cordon or uncordon node from k8s cluster.
#141 

##### ISSUE TYPE

- New Module Pull Request

##### COMPONENT NAME

k8s_drain


plugins/modules/k8s_drain.py
```diff
@@ -0,0 +1,415 @@
...
+from __future__ import absolute_import, division, print_function
+
+__metaclass__ = type
+
+
+DOCUMENTATION = r'''
+
+module: k8s_drain
+
+short_description: Drain, Cordon, or Uncordon node in k8s cluster.
```
Akasurde, line 16:
```suggestion
short_description: Drain, Cordon, or Uncordon node in k8s cluster
```

plugins/modules/k8s_drain.py
```diff
@@ -0,0 +1,415 @@
...
+
+author: Aubin Bikouo (@abikouo)
+
+description:
+  - Drain node in preparation for maintenance same as kubectl drain.
+  - Cordon will mark node as unschedulable.
+  - Uncordon will mark node as schedulable.
+  - The given node will be marked unschedulable to prevent new pods from arriving.
+  - Then drain deletes all pods except mirror pods (which cannot be deleted through the API server).
+
```
Akasurde, line 26:
```suggestion
  - Drain node in preparation for maintenance same as kubectl drain.
  - Cordon will mark the node as unschedulable.
  - Uncordon will mark the node as schedulable.
  - The given node will be marked unschedulable to prevent new pods from arriving.
  - Then drain deletes all pods except mirror pods (which cannot be deleted through the API server).

```

plugins/modules/k8s_drain.py
```diff
@@ -0,0 +1,415 @@
...
+extends_documentation_fragment:
+  - kubernetes.core.k8s_auth_options
+
+options:
+  state:
+    description:
+    - Determines whether to drain, cordon, or uncordon node.
```
Akasurde, line 33: Can we name this parameter something else, since `state` is known for values like `present` and `absent`? Something like `operation`.

plugins/modules/k8s_drain.py
```diff
@@ -0,0 +1,415 @@
...
+            type: bool
+            default: False
+        disable_eviction:
+            description:
+            - forces drain to use delete rather than evict.
+            type: bool
+            default: False
+        wait_timeout:
+            description:
+            - The length of time to wait for pod to be delected/evicted before giving up, zero means infinite.
```
Akasurde, line 72: Unit for time? seconds, minutes?

plugins/modules/k8s_drain.py
```diff
@@ -0,0 +1,415 @@
...
+            type: bool
+            default: False
+        ignore_daemonsets:
+            description:
+            - Ignore DaemonSet-managed pods.
+            type: bool
+            default: False
+        disable_eviction:
+            description:
+            - forces drain to use delete rather than evict.
```
Akasurde, line 67:
```suggestion
            - Forces drain to use delete rather than evict.
```

plugins/modules/k8s_drain.py
```diff
@@ -0,0 +1,415 @@
...
+            type: int
+
+requirements:
+  - python >= 3.6
+  - kubernetes >= 12.0.0
+'''
+
+EXAMPLES = r'''
+- name: Drain node "foo", even if there are pods not managed by a ReplicationController, Job, or DaemonSet on it.
+  kubernetes.core.k8s_node:
```
Akasurde, line 83:
```suggestion
  kubernetes.core.k8s_drain:
```

plugins/modules/k8s_drain.py
abikouo: seconds, thx

plugins/modules/k8s_drain.py
```diff
@@ -0,0 +1,415 @@
...
+        response = None
+        pod = pods.pop()
+        while (_elapsed_time() < wait_timeout or wait_timeout == 0) and pods:
+            if not pod:
+                pod = pods.pop()
+            try:
+                response = self._api_instance.read_namespaced_pod(namespace=pod[0], name=pod[1])
+                if not response:
+                    pod = None
+                time.sleep(1)
```
Akasurde, line 240: Not a blocker, but can we make this sleep configurable.

plugins/modules/k8s_drain.py
```diff
@@ -0,0 +1,415 @@
...
+            except ApiException as exc:
+                if exc.reason != "Not Found":
+                    self._module.fail_json(msg="Exception raised: {0}".format(exc.reason))
+                pod = None
+            except Exception as e:
+                self._module.fail_json(msg="Exception raised: {0}".format(to_native(e)))
+        if not pods:
+            return None
+        else:
+            return "timeout reached while pods were still running."
```
Akasurde, line 250:
```suggestion
        if not pods:
            return None
        return "timeout reached while pods were still running."
```

plugins/modules/k8s_drain.py
```diff
@@ -0,0 +1,415 @@
...
+        if warnings:
+            return dict(result=' '.join(result), warnings=warnings)
+        return dict(result=' '.join(result))
+
+    def patch_node(self, unschedulable):
+        try:
+            body = {
+                'spec': {'unschedulable': unschedulable}
+            }
+            self._api_instance.patch_node(name=self._module.params.get('name'), body=body)
```
Akasurde, line 323:
```suggestion

        body = {
            'spec': {'unschedulable': unschedulable}
        }
        try:
            self._api_instance.patch_node(name=self._module.params.get('name'), body=body)
```

plugins/modules/k8s_drain.py
abikouo: the ``state`` is used for some modules like the ``ansible.builtin.file`` module with value different from ``present`` and ``absent``, but here of course this can be changed if confusing

plugins/modules/k8s_drain.py
```diff
@@ -0,0 +1,394 @@
...
+        self._k8s_ansible_mixin.argspec = self._module.argument_spec
+        self._k8s_ansible_mixin.check_mode = self._module.check_mode
+        self._k8s_ansible_mixin.params = self._module.params
+        self._k8s_ansible_mixin.fail_json = self._module.fail_json
+        self._k8s_ansible_mixin.fail = self._module.fail_json
+        self._k8s_ansible_mixin.exit_json = self._module.exit_json
+        self._k8s_ansible_mixin.warn = self._module.warn
+        self._k8s_ansible_mixin.warnings = []
+
+        self._api_instance = core_v1_api.CoreV1Api(self._k8s_ansible_mixin.client.client)
```
gravesm, line 211: Is there a reason we aren't using the dynamic client here? I'd prefer to keep things consistent throughout the code base even if it means a few extra lines of code.

abikouo: dynamic cannot be used here as it does not return owner references (for pod) which is mandatory to determine whether a pod is eligible for deletion/eviction. Also, we are working exclusively with the ``pod`` object so so there is no added value to use the dynamic client.

abikouo:
recheck

goneri:
recheck

plugins/modules/k8s_drain.py
```diff
@@ -0,0 +1,394 @@
...
+                self._module.fail_json(msg="Failed to list pod from node {name} due to: {reason}".format(
+                                       name=self._module.params.get('name'), reason=exc.reason), status=exc.status)
+            pods = []
+        except Exception as exc:
+            self._module.fail_json(msg="Failed to list pod from node {name} due to: {error}".format(
+                                   name=self._module.params.get('name'), error=to_native(exc)))
+
+        # Mark node as unschedulable
+        result = []
+        if not node_unschedulable:
```
gravesm, line 296: This needs to come before you get the pod list, otherwise there's a race condition.

plugins/modules/k8s_drain.py
```diff
@@ -0,0 +1,394 @@
...
+            self.patch_node(unschedulable=True)
+            result.append("node {0} marked unschedulable.".format(self._module.params.get('name')))
+            self._changed = True
+        else:
+            warnings.append("node {0} already marked unschedulable.".format(self._module.params.get('name')))
```
gravesm, line 301: I'm not sure this needs to be a warning. I'd just append it to the result.

plugins/modules/k8s_drain.py
```diff
@@ -0,0 +1,394 @@
...
+
+        # Delete Pods
+        if pods:
+            self._changed = True
```
gravesm, line 305: This won't necessarily be true. Since this is set correctly in the `evict_pods` method, I would delete it here.

molecule/default/tasks/drain.yml
```diff
@@ -0,0 +1,212 @@
...
+        ignore_daemonsets: true
+        wait_timeout: 0
+    register: drain_result
+
+  - name: assert that node has been drained
+    assert:
+      that:
+      - drain_result is changed
+      - '"node {{ node_to_drain }} marked unschedulable." in drain_result.result'
+
```
gravesm, line 175: Probably worth adding a test to make sure the unmanaged pod was deleted.

gravesm (changes requested):

gravesm (approved):

Akasurde (approved):
LGTM

PR #144, opened 2021-06-22
molecule gc.yml - fix sporadic fail

abikouo:
##### SUMMARY

Improve https://github.com/ansible-collections/kubernetes.core/pull/132

### DIFF
--- a/molecule/default/tasks/gc.yml
+++ b/molecule/default/tasks/gc.yml
@@ -37,18 +37,25 @@
       k8s:
         definition: "{{ job_definition }}"
 
-    - name: Test that job's pod is running
+    - name: Wait Job's pod
       k8s_info:
         kind: Pod
         namespace: "{{ gc_namespace }}"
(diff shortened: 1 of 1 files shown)


gravesm (commented):
Does the `until` fix not work?

Akasurde:
> Does the `until` fix not work?

I guess no. Since it is not able to find `status` in `resource[0]`

abikouo:
> > Does the `until` fix not work?
> 
> I guess no. Since it is not able to find `status` in `resource[0]`

indeed seems not enough, and we cannot add a wait condition for Job resource type as the API does not return information on Pods statuses.

Akasurde:
@abikouo Do you know any other predictive way by which we can remove the hardcoded sleep value?

Akasurde:
Make sense to me.

gravesm (approved):
I will be glad to have this fixed. :tada:

Issue #88, opened 2021-04-28
`diff: no` is not respected with k8s


benjdudas:
##### SUMMARY

When running a playbook with the diff option,  diff results are being listed when `diff: no` is  configured for the task. 

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

community.kubernetes.k8s

##### ANSIBLE VERSION

```paste below
ansible 2.10.8
  config file = /etc/ansible/ansible.cfg
  configured module search path = ['/home/bdudas/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/bdudas/dev-net/ansible_2_10/lib64/python3.9/site-packages/ansible
  executable location = /home/bdudas/dev-net/ansible_2_10/bin/ansible
  python version = 3.9.2 (default, Feb 20 2021, 00:00:00) [GCC 10.2.1 20201125 (Red Hat 10.2.1-9)]
```
```
Collection           Version
-------------------- -------
community.kubernetes 1.2.1
```

##### CONFIGURATION

```paste below
ANSIBLE_COW_SELECTION(/etc/ansible/ansible.cfg) = www
HOST_KEY_CHECKING(/etc/ansible/ansible.cfg) = False
RETRY_FILES_ENABLED(/etc/ansible/ansible.cfg) = False
```

##### STEPS TO REPRODUCE

```yaml
    - name: test diff on kube definition
      community.kubernetes.k8s:
        state: present
        apply: yes
        namespace: default
        definition: "{{ lookup('file', 'awx-secret.yml') }}"
        kubeconfig: "{{ kube_location }}"
      register: results
      diff: no

```

##### EXPECTED RESULTS

Diff results will be disabled for the task. 

https://docs.ansible.com/ansible/latest/user_guide/playbooks_checkmode.html#using-diff-mode

##### ACTUAL RESULTS

Diff results for the task continue to be displayed in stdout

Akasurde:
@benjdudas Could you please check if PR #146 works for you and let us know? Thanks.

Akasurde:
resolved_by_pr #146

abikouo:
Feel free to re-open the issue if needed

PR #146, opened 2021-06-23
k8s support diff mode

abikouo:
##### SUMMARY

``diff`` attribute was always returned in output when ``k8s`` module was executed using options, ``force``, ``apply`` or ``state=patched`` even if the ``diff=no`` was set at task level.
This fix aims to correct this
#88 

This fix is not working if collection ``cloud.common`` is used without the fix https://github.com/ansible-collections/cloud.common/pull/65

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

k8s

### DIFF
--- a/changelogs/fragments/146-k8s-add-support-diff-mode.yml
+++ b/changelogs/fragments/146-k8s-add-support-diff-mode.yml
@@ -1,2 +1,2 @@
-breaking_changes:
+bugfixes:
   - k8s - Add support for diff mode, diff is returned only when it is explicitly requested (https://github.com/ansible-collections/kubernetes.core/pull/146).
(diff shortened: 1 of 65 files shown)


changelogs/fragments/146-k8s-add-support-diff-mode.yml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - k8s - Add support for diff mode, diff is returned only when it is explicitly requested (https://github.com/ansible-collections/kubernetes.core/pull/146).
```
gravesm, line 2:
```suggestion
  - k8s - Fixes a bug where diff was always returned when using apply or modifying an existing object, even when diff=no was specified. The module no longer returns diff unless requested and will now honor diff=no (https://github.com/ansible-collections/kubernetes.core/pull/146).
```

gravesm (changes requested):
Just realized we also need to fix the `k8s_scale` and `k8s_json_patch` modules, too.

gravesm:
> Just realized we also need to fix the k8s_scale and k8s_json_patch modules, too.

This still needs to be addressed.

abikouo:
> > Just realized we also need to fix the k8s_scale and k8s_json_patch modules, too.
> 
> This still needs to be addressed.

done

gravesm (approved):
Thanks!

PR #151, opened 2021-06-24
Check that auth value is not None

gravesm:
##### SUMMARY

The previous check for truth prevented the verify_ssl param from being
set to false, thus forcing ssl verfication in every case.

##### ISSUE TYPE

- Bugfix Pull Request

##### ADDITIONAL INFORMATION

```paste below

```


jillr (approved):
lgtm

Akasurde (approved):

pabelanger:
recheck

PR #153, opened 2021-06-25
Improve CI running time

abikouo:
##### SUMMARY
Molecule default scenario will be split into multiple github jobs based on the tags.
We are a separate job for the lint. This could be merged once the following is merge into molecule repository: https://github.com/ansible-community/molecule/pull/3166
##### ISSUE TYPE

- Feature Pull Request

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -70,12 +70,11 @@ jobs:
         with:
           fail_ci_if_error: false
 
-  molecule:
+  molecule-lint:
     runs-on: ubuntu-latest
     strategy:
       matrix:
         python_version: ['3.7']
-        ansible_version: ['==2.9.*', '==2.10.*', '']
     steps:
       - name: Check out code
         uses: actions/checkout@v2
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -70,12 +70,59 @@ jobs:
...
+
+      - name: Set up KinD cluster
+        uses: engineerd/setup-kind@v0.5.0
+
+      - name: Set up Python ${{ matrix.python_version }}
+        uses: actions/setup-python@v1
+        with:
+          python-version: ${{ matrix.python_version }}
+
+      # The 3.3.0 release of molecule introduced a breaking change. See
```
gravesm, line 92: I think this has actually been fixed. We should confirm, but we can probably remove this restriction.

Makefile
```diff
@@ -24,8 +25,11 @@ test-sanity:
 test-integration:
 	ansible-test integration --docker -v --color --retry-on-error --python $(PYTHON_VERSION) --continue-on-error --diff --coverage $(?TEST_ARGS)
 
-test-molecule:
-	molecule test
+converge-molecule:
+	molecule converge -- -v --tags $(ANSIBLE_TAGS)
```
gravesm, line 29: Can we parameterize the verbosity here, defaulting to `-v`?

gravesm (commented):
Thanks for the fix upstream in molecule. My only suggestion would be that we might want to group some of the tags together rather than having everything run separately. That's a lot of checks. Probably quite a few of the faster k8s tags can all run together in the same time it takes the helm tests to run.

Makefile
abikouo: yes can do. Thanks for the feedback

abikouo:
> Thanks for the fix upstream in molecule. My only suggestion would be that we might want to group some of the tags together rather than having everything run separately. That's a lot of checks. Probably quite a few of the faster k8s tags can all run together in the same time it takes the helm tests to run.

Yes I will do, after the first run I will make some time analysis to determine which can be grouped

tadeboro:
I know no one asked for my opinion, but here it is anyway ;) Is there a reason why there is only one molecule scenario used to test the collection? From my experience, messing with tags is really error-prone and can quickly lead to some really hard-to-debug issues.

We use molecule a lot and we learned (the hard way, as always) that having tests split over multiple scenarios works best. For example, in the `sensu.sensu_go` collection I help maintain, we have around 40 scenarios: one for each module pair (module + module_info) and at least one scenario for each role we provide. This way, we can easily run only part of the test suite.

And to keep the test times (relatively) short, we then distribute the scenarios across a few executors. At the start, we statically distributed those scenarios into different groups based on some local timing information and things worked really nice even then. So maybe that would be something this collection can also use? If there is interest, I can provide some links to our setup at that time.

(And if you are really interested in some funky CI/CD automation, our current setup is probably the place to go since we use pytest to run and time our scenario executions. This timing is then used to automatically partition scenarios in future CI/CD runs. CircleCI is really awesome in this respect.)

gravesm:
@tadeboro Thanks for the advice! I would say the reason we are using one scenario is mostly historical. @abikouo did have a PR that split things out into multiple scenarios, though using tags seems like less work. We have not seen any issues yet, but I'm sure you have more experience with molecule than I do. I'd be interested in seeing what you've done, particularly around balancing scenarios across a test run.

abikouo:
no more relevant, will use a new strategy as we migrate to zuul

Issue #147, opened 2021-06-23
lookup and query behavior differ when using resource_name vs label_selector


jmazzitelli:
Run this playbook:

```
- hosts: localhost
  gather_facts: no
  vars:
    ansible_python_interpreter: python3
  collections:
  - kubernetes.core
  tasks:
  - k8s:
      state: present
      definition:
        apiVersion: v1
        kind: Namespace
        metadata:
          name: testns
          labels:
            customLabel: test
  - assert:
      that: query('kubernetes.core.k8s', kind='Namespace', resource_name='testns')[0].metadata.name == "testns"
  - assert:
      that: query('kubernetes.core.k8s', kind='Namespace', label_selector='customLabel=test')[0].metadata.name == "testns"
    ignore_errors: yes

  - assert:
      that: lookup('kubernetes.core.k8s', kind='Namespace', resource_name='testns')[0].metadata.name == "testns"
    ignore_errors: yes
  - assert:
      that: lookup('kubernetes.core.k8s', kind='Namespace', label_selector='customLabel=test')[0].metadata.name == "testns"
```

Here I have a pair of `query` and a pair of `lookup` invocations. In each pair, one is looking up the namespace by name (using `resource_name`) and one is looking up the namespace by label (using `label_selector`).

Each assert is verifying the same thing - that the first element in the returned list has `metadata.name == "testns"` (thus verifying the results include the namespace resource in the first element of the list).

The asserts with "ignore_errors: yes" are the ones that are failing (but I don't think they should be failing).

When using `query`, looking up by resource_name is OK, but an error occurs using label_selector.
It is the reverse when using `lookup` (looking up by label_selector is OK, but an error occurs using resource_name).

Here's the full results output:

```
PLAY [localhost] ***********************************************************************************************************************

TASK [k8s] *****************************************************************************************************************************
ok: [localhost]

TASK [assert] **************************************************************************************************************************
ok: [localhost] => {
    "changed": false,
    "msg": "All assertions passed"
}

TASK [assert] **************************************************************************************************************************
fatal: [localhost]: FAILED! => {"msg": "The conditional check 'query('kubernetes.core.k8s', kind='Namespace', label_selector='customLabel=test')[0].metadata.name == \"testns\"' failed. The error was: error while evaluating conditional (query('kubernetes.core.k8s', kind='Namespace', label_selector='customLabel=test')[0].metadata.name == \"testns\"): 'list object' has no attribute 'metadata'"}
...ignoring

[...]

PR #155, opened 2021-06-28
lookup: Recommend query instead of lookup

Akasurde:
##### SUMMARY

To provide consistent interface for k8s lookup plugin,
add recommendation for `query` or `lookup` with `wantlist=True`.

Fixes: #147

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
changelogs/fragments/147_lookup.yml
molecule/default/tasks/lookup_k8s.yml
plugins/lookup/k8s.py

### DIFF
--- a/changelogs/fragments/147_lookup.yml
+++ b/changelogs/fragments/147_lookup.yml
@@ -1,0 +1,3 @@
+---
+bugfixes:
+- lookup - recommend query instead of lookup (https://github.com/ansible-collections/kubernetes.core/issues/147).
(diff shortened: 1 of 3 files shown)


changelogs/fragments/147_lookup.yml
```diff
@@ -0,0 +1,3 @@
+---
+minor_changes:
```
gravesm, line 2: Should this be `bugfixes`? I feel like if it's not behaving the way the ansible documentation says it should behave then it's a bug.

Akasurde: Got it. I am checking `errors=ignore` behavior and will update with that.

gravesm (approved):

PR #158, opened 2021-06-30
k8s - add label_selectors options

### LINKED ISSUE #43: Support the selector option in the k8s module when applying service definitions
##### SUMMARY

This is a copy of the following GH issue: https://github.com/ansible/ansible/issues/67316.

The k8s module should support specifying selectors when applying resource definitions, similar to how `kubectl` supports it.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
k8s module

##### ADDITIONAL INFORMATION

Some commonly used projects, such as Knative, require you to apply the Custom Resource Definitions (CRDs) to the K8s cluster, before you can deploy the rest of the service definition.

See: https://knative.dev/v0.11-docs/install/knative-with-any-k8s/#installing-knative.

Without the support for specifying selectors, this is not possible in Ansible without having to use a 3rd party binary (`kubectl`). 

```yaml

# example 
- name: Install Knative CRDs                                                                                                                                                                                      
  k8s:                                   
    state: present
    src: "{{ item.dest }}"
    context: "{{ k8s_cluster | mandatory }}" 
    # this does not work
   selector:
     - 'knative.dev/crd-install=true'    
    validate:
      fail_on_error: yes                                                                                                                                                                                   
  with_items:
    - "{{ knative_install_files.results }}"

# output

msg: 'Unsupported parameters for (k8s) module: selector Supported parameters include: api_key, api_version, append_hash, apply, ca_cert, client_cert, client_key, context, force, host, kind, kubeconfig, merge_type, name, namespace, password, proxy, resource_definition, src, state, username, validate, validate_certs, wait, wait_condition, wait_sleep, wait_timeout'
```

Sample `kubectl` command utilizing selectors:

```bash
kubectl apply --selector knative.dev/crd-install=true \
--filename https://github.com/knative/serving/releases/download/v0.11.0/serving.yaml \
--filename https://github.com/knative/eventing/releases/download/v0.11.0/release.yaml \
--filename https://github.com/knative/serving/releases/download/v0.11.0/monitoring.yaml
```

abikouo:
##### SUMMARY
``k8s`` now support ``label_selectors`` options same as ``k8s_info``

Resolves #43 

##### ISSUE TYPE

- Feature Pull Request
- 
##### COMPONENT NAME

k8s


plugins/module_utils/common.py
```diff
@@ -663,6 +677,14 @@ def build_error_msg(kind, name, msg):
                 return result
 
         else:
+            if label_selectors:
+                filter_selector = LabelSelectorFilter(label_selectors)
+                if not filter_selector.isMatching(definition):
+                    result['changed'] = False
+                    del result['result']
+                    result['warning'] = "resource 'kind={kind},name={name},namespace={namespace}' filtered by label_selectors.".format(
```
gravesm, line 685: This isn't really a warning because it's exactly what the user requested. I'd suggest we instead use `msg` here.

plugins/module_utils/common.py
```diff
@@ -663,6 +677,14 @@ def build_error_msg(kind, name, msg):
                 return result
 
         else:
+            if label_selectors:
+                filter_selector = LabelSelectorFilter(label_selectors)
+                if not filter_selector.isMatching(definition):
+                    result['changed'] = False
+                    del result['result']
```
gravesm, line 684: If we delete the result key, this is a breaking change. It should just be an empty dict.

gravesm (changes requested):
The label_selector on the k8s module should support the same functionality as the label_selector on the k8s_info module, otherwise, it's going to be confusing. Specifically, we need to add set support for set-based requirements. See https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#label-selectors.

abikouo:
> The label_selector on the k8s module should support the same functionality as the label_selector on the k8s_info module, otherwise, it's going to be confusing. Specifically, we need to add support for set-based requirements. See https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#label-selectors.

ok this is not supporting set-based requirement now as there are some filters performed locally, it would be great to have this directly done on kubernetes python client. I will work on it

abikouo:
recheck

plugins/module_utils/selector.py
```diff
@@ -0,0 +1,96 @@
...
+# limitations under the License.
+
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+import re
+
+
+class Selector(object):
```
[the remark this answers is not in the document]
abikouo: Thats true, will try this
Thanks

gravesm (approved):

PR #160, opened 2021-07-01
Helm tests failing on CI - Fix

abikouo:
##### SUMMARY

Fix molecule test Helm - unexpected failure

##### ISSUE TYPE

- Bugfix Pull Request
##### COMPONENT NAME

Molecule test - Helm

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -82,8 +82,11 @@ jobs:
         with:
           path: ansible_collections/kubernetes/core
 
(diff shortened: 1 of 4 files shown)


gravesm (changes requested):
Thanks for figuring this out. Could you add a comment, maybe in the prepare.yml file, about why this change was made? This is a pretty esoteric change and I'd like to make sure we can look back and understand why it was done.

gravesm (approved):

PR #169, opened 2021-07-15
Re-enable support for turbo mode

gravesm:
##### SUMMARY

This re-enables the ability to add turbo mode. It also adds a few more
tests to cover some cases that had been broken in turbo mode previously.

Testing with turbo mode is not currently enabled, and would fail until https://github.com/ansible-collections/cloud.common/pull/69 can be merged and a new cloud.common release is done. This also does not add cloud.common to the collection dependencies until a decision has been made about how enabling/disabling turbo mode will work when cloud.common is already installed.

##### ISSUE TYPE

- Bugfix Pull Request

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/plugins/module_utils/ansiblemodule.py
+++ b/plugins/module_utils/ansiblemodule.py
@@ -3,10 +3,22 @@ from __future__ import (absolute_import, division, print_function)
 __metaclass__ = type
 
 
+import os
+
+from ansible.module_utils.common.validation import check_type_bool
+
 try:
-    from ansible_collections.cloud.common.plugins.module_utils.turbo.module import (
-        AnsibleTurboModule as AnsibleModule,
-    )  # noqa: F401
-    AnsibleModule.collection_name = "kubernetes.core"
-except ImportError:
+    enable_turbo_mode = check_type_bool(os.environ.get("ENABLE_TURBO_MODE"))
+except TypeError:
+    enable_turbo_mode = False
+
+if enable_turbo_mode:
+    try:
+        from ansible_collections.cloud.common.plugins.module_utils.turbo.module import (
+            AnsibleTurboModule as AnsibleModule,
+        )  # noqa: F401
+        AnsibleModule.collection_name = "kubernetes.core"
+    except ImportError:
+        from ansible.module_utils.basic import AnsibleModule  # noqa: F401
+else:
     from ansible.module_utils.basic import AnsibleModule  # noqa: F401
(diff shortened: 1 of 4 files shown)


plugins/module_utils/ansiblemodule.py
```diff
@@ -3,4 +3,10 @@
 __metaclass__ = type
 
 
-from ansible.module_utils.basic import AnsibleModule  # noqa: F401
+try:
+    from ansible_collections.cloud.common.plugins.module_utils.turbo.module import (
+        AnsibleTurboModule as AnsibleModule,
+    )  # noqa: F401
+    AnsibleModule.collection_name = "kubernetes.core"
+except ImportError:
```
Akasurde, line 11: LGTM, are we waiting for any additional changes.

gravesm: https://github.com/ansible-collections/cloud.common/pull/69 needs to be merged and released before the tests here will pass with turbo mode enabled. If the turbo mode tests are non-voting then we can merge and just have broken tests until the issue with cloud.common is resolved.

gravesm:
@Akasurde if you want to give the new changes a review, I think we can move forward with this for now and adjust later if we need to.

Akasurde (approved):
LGTM. I hope we are adding docs are around this env variable in the later PR.

gravesm:
Yeah, I would like to get turbo mode settled and make sure it's being tested before we publish docs for it.

gravesm:
To clarify, what I'm suggesting is that the existing `environment` keyword be used to control this:
```
- k8s:
    definition:
       ...
  environment:
    ENABLE_TURBO_MODE: true
```

goneri:
Oh crap, I totally missed this PR and I started working on this in parallel: https://github.com/ansible-collections/kubernetes.core/pull/190 and https://github.com/ansible-collections/cloud.common/pull/73

gravesm:
@goneri Yeah, I think I made this after yours as an alternative to consider. I'm not sure which one is better. Mine seems simpler but I may be missing some reason why we wouldn't want to do things this way.

goneri:
> @goneri Yeah, I think I made this after yours as an alternative to consider. I'm not sure which one is better. Mine seems simpler but I may be missing some reason why we wouldn't want to do things this way.

Yeah. I need to think about this a bit. You're approach is definitely way lighter.

gravesm:
I believe it should also be easy to set up a non-voting CI job using something like:

```
environment:
  ENABLE_TURBO_MODE: "{{ lookup('env', 'ENABLE_TURBO_MODE') }}"
```
at the playbook level in `converge.yml`.

goneri (approved):

goneri:
I suggest we land your PR first. And in the future, if we look for something more elaborate, we can still revisit mine.

PR #173, opened 2021-07-19
Add script to handle sanity ignore

Akasurde:
Depends-on: https://github.com/ansible/ansible-zuul-jobs/pull/993

##### SUMMARY

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
tests/sanity/ignore-2.10.txt
tests/sanity/ignore-2.11.txt
tests/sanity/ignore-2.12.txt
tests/sanity/ignore-2.9.txt
tests/sanity/refresh_ignore_files


tests/sanity/refresh_ignore_files
```diff
@@ -0,0 +1,42 @@
...
+for version in ["2.9", "2.10", "2.11", "2.12"]:
+    ignore_file = ignore_dir / f"ignore-{version}.txt"
+    ignore_content = ignore_file.read_text().split("\n")
+    ignore_content.append(f"tests/sanity/refresh_ignore_files shebang!skip")
+    skip_list = [
+        "compile-2.6!skip",  # Py3.8+
+        "compile-2.7!skip",  # Py3.8+
+        "compile-3.5!skip",  # Py3.8+
+        "compile-3.6!skip",  # Py3.8+
+        "compile-3.7!skip",  # Py3.8+
```
gravesm, line 21: Do we even need lines for < 3.6? We are only testing on 3.6+. I think I'd also prefer we don't skip the compile check on versions we support. I don't think we've had issues in the past with this check. I would say just remove compile from this list.

tests/sanity/refresh_ignore_files
```diff
@@ -0,0 +1,42 @@
...
+        "import-2.6!skip",  # Py3.8+
+        "import-2.7!skip",  # Py3.8+
+        "import-3.5!skip",  # Py3.8+
+        "import-3.6!skip",  # Py3.8+
+        "import-3.7!skip",  # Py3.8+
```
gravesm, line 26: Same as above, do we need this for < 3.6? I'm also a little hesitant to disable the import check on everything. We only have a few files where this is needed. I feel like a better solution would be to manage a list of files in this script we want to skip import checks on and add the skip based on that for all ansible versions.

tests/sanity/refresh_ignore_files
Akasurde: Zuul runs all the versions of Python for sanity tests. There is no simple way to disable these checks unless we add these lines for non-supported versions. This is already been followed in `amazon.aws` repo.

goneri (approved):

tests/sanity/refresh_ignore_files
```diff
@@ -0,0 +1,39 @@
...
+            continue
+        for test in skip_list:
+            ignore_content.append(f"{f} {test}")
+    for f in module_utils_dir.glob("*.py"):
+        if f.is_symlink():
+            continue
+        for test in skip_list:
+            ignore_content.append(f"{f} {test}")
+    ignore_file = ignore_dir / f"ignore-{version}.txt"
+    ignore_file.write_text("\n".join(sorted(set(ignore_content))).lstrip("\n"))
```
abikouo, line 39:
```suggestion
#!/usr/bin/env python3

from pathlib import Path
import argparse

if __name__ == "__main__":
    parser = argparse.ArgumentParser(description="Refresh ignore files for sanity tests.")
    parser.add_argument("--ansible-release", default="2.9,2.10,2.11,2.12", dest="release", help="comma-separated list of ansible release version")
    args = parser.parse_args()
    target_dir = Path('.')

    ignore_dir = target_dir / "tests" / "sanity"
    module_dir = target_dir / "plugins" / "modules"
    module_utils_dir = target_dir / "plugins" / "module_utils"
    ignore_dir.mkdir(parents=True, exist_ok=True)

    skip_list = [
        "compile-2.6!skip",  # Py3.8+
        "compile-2.7!skip",  # Py3.8+
        "compile-3.5!skip",  # Py3.8+
        "import-2.6!skip",  # Py3.8+
        "import-2.7!skip",  # Py3.8+
        "import-3.5!skip",  # Py3.8+
        "future-import-boilerplate!skip",  # Py2 only
        "metaclass-boilerplate!skip",  # Py2 only
    ]

    for version in [ x.replace(" ", "") for x in args.release.split(",")]:
        ignore_file = ignore_dir / f"ignore-{version}.txt"
        ignore_content = ignore_file.read_text().split("\n")
        ignore_content.append(f"tests/sanity/refresh_ignore_files shebang!skip")

        for f in module_dir.glob("*.py"):
            if f.is_symlink():
                continue
            for test in skip_list:
                ignore_content.append(f"{f} {test}")
        for f in module_utils_dir.glob("*.py"):
            if f.is_symlink():
                continue
            for test in skip_list:
                ignore_content.append(f"{f} {test}")
        ignore_file = ignore_dir / f"ignore-{version}.txt"
        ignore_file.write_text("\n".join(sorted(set(ignore_content))).lstrip("\n"))
```

abikouo: This is basically a way to add configuration on script in order to support future ansible release.

PR #190, opened 2021-07-28
allow turbo mode, but disabled by default

goneri:
Depends-On: https://github.com/ansible-collections/cloud.common/pull/73

Expose a new parameter called enable_session_cache to enable the
feature.

### DIFF
--- a/changelogs/fragments/add_enable_session_cache.yaml
+++ b/changelogs/fragments/add_enable_session_cache.yaml
@@ -1,0 +1,3 @@
+---
+minor_changes:
+- add ``enable_session_cache`` parameter to turn on the session cache. The option is disabled by default.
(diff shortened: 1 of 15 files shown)


gravesm:
This option will need to be added to most of the k8s_* modules.

goneri:
> This option will need to be added to most of the k8s_* modules.

Yes, I'm doing my test with just k8s for now. @gravesm  How do you feel with this design?

gravesm:
I think an env var or a global ansible setting would be a better solution, but I understand those will be much more difficult to implement. I'll do some testing on my own end to see how this works.

goneri:
Superseded by https://github.com/ansible-collections/kubernetes.core/pull/169

PR #204, opened 2021-08-16
helm: List all matching releases, regardless of state

RuriRyan:
##### SUMMARY
There are cases where a `helm list` does not list all
existing releases. One such case is when a release
is in state pending-upgrade, or any other pending state.

Now with this change any use of the lookup or info modules
will now return such releases, so that one can work with them.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
`kuberntes.core.helm` and `kubernetes.core.helm_info`

##### ADDITIONAL INFORMATION
Before:
```
➜  kubernetes.core git:(main) helm list -n kube-system --output=yaml --filter release-name                      
[]
```
After:
```
➜  kubernetes.core git:(main) helm list -a -n kube-system --output=yaml --filter release-name                      
- app_version: v1.3.4
  chart: release-name-1.3.9
  name: release-name
  namespace: kube-system
  revision: "47"
  status: pending-upgrade
  updated: 2021-08-16 13:51:18.713253195 +0200 CEST
```

As the output was completely missing before it is now shown
and can be parsed and processed accordingly.

### DIFF
--- a/changelogs/fragments/204_helm-list-all.yaml
+++ b/changelogs/fragments/204_helm-list-all.yaml
@@ -1,0 +1,3 @@
+---
+bugfixes:
+  - helm - lookup and info modules now return all releases regardless of their state (https://github.com/ansible-collections/kubernetes.core/pull/204).
--- a/molecule/default/roles/helm/tasks/tests_chart.yml
+++ b/molecule/default/roles/helm/tasks/tests_chart.yml
@@ -250,7 +250,7 @@
         binary_path: "{{ helm_binary }}"
         state: absent
         name: test-0001
-        purge: False
+        purge: True
         namespace: "{{ helm_namespace }}"
       register: install
 
--- a/plugins/modules/helm.py
+++ b/plugins/modules/helm.py
@@ -330,7 +330,7 @@ def get_release_status(module, command, release_name):
     Get Release state from deployed release
     """
 
-    list_command = command + " list --output=yaml --filter " + release_name
+    list_command = command + " list -a --output=yaml --filter " + release_name
 
     rc, out, err = run_helm(module, list_command)
 
--- a/plugins/modules/helm_info.py
+++ b/plugins/modules/helm_info.py
@@ -113,7 +113,7 @@ def get_release(state, release_name):
 
 # Get Release state from deployed release
 def get_release_status(module, command, release_name):
-    list_command = command + " list --output=yaml --filter " + release_name
+    list_command = command + " list -a --output=yaml --filter " + release_name
 
     rc, out, err = run_helm(module, list_command)
 


abikouo:
@RuriRyan Thanks for the PR.
Could you also provide a changelog fragment? https://docs.ansible.com/ansible/latest/community/development_process.html#community-changelogs

abikouo:
recheck

abikouo (approved):
LGTM!

gravesm:
@RuriRyan Thanks for the PR. I think it would definitely be a good idea to add this sort of functionality to `helm_info`. We should probably do this by adding a new boolean parameter to the module that controls this behavior, defaulting to false so we aren't potentially breaking any existing playbooks. This would also mirror the default behavior of `helm list`.

I am not so sure about making this change for the `helm` module, though. This change in the helm module breaks multiple existing behaviors. Consider the following playbook:
```
- kubernetes.core.helm:
    state: absent
    name: foo
    purge: false
    namespace: foobar

- kubernetes.core.helm:
    state: absent
    name: foo
    purge: false
    namespace: foobar
```

This works before your change, but fails on the second task after with an error message that foo is already deleted.

In this playbook:

```
- kubernetes.core.helm:
    state: absent
    name: foo
    purge: false
    namespace: foobar

- kubernetes.core.helm:
    name: foo
    namespace: foobar
    replace: true
    chart_ref: ./foo
``` 
Before your change, the end result is a `foo` release that has the status `deployed`. After your change, the end result is a `foo` release that has the status `uninstalled`. These problems are mostly because we aren't examining the status of a release. It may be that we want to revisit how we are handling this in the `helm` module, but that feels like a different (though related) issue from a change to `helm_info`. I would rather address any change to the `helm` module in a separate issue after considering what the impact would be and what the desired behavior should be.

abikouo:
> @RuriRyan Thanks for the PR. I think it would definitely be a good idea to add this sort of functionality to `helm_info`. We should probably do this by adding a new boolean parameter to the module that controls this behavior, defaulting to false so we aren't potentially breaking any existing playbooks. This would also mirror the default behavior of `helm list`.
> 
> I am not so sure about making this change for the `helm` module, though. This change in the helm module breaks multiple existing behaviors. Consider the following playbook:
> 
> ```
> - kubernetes.core.helm:
>     state: absent
>     name: foo
>     purge: false
>     namespace: foobar
> 
> - kubernetes.core.helm:
>     state: absent
>     name: foo
>     purge: false
>     namespace: foobar
> ```
> 
> This works before your change, but fails on the second task after with an error message that foo is already deleted.
> 
> In this playbook:
> 
> ```
> - kubernetes.core.helm:
>     state: absent
>     name: foo
>     purge: false
>     namespace: foobar
> 
> - kubernetes.core.helm:
>     name: foo
>     namespace: foobar
>     replace: true
>     chart_ref: ./foo
> ```
> 
> Before your change, the end result is a `foo` release that has the status `deployed`. After your change, the end result is a `foo` release that has the status `uninstalled`. These problems are mostly because we aren't examining the status of a release. It may be that we want to revisit how we are handling this in the `helm` module, but that feels like a different (though related) issue from a change to `helm_info`. I would rather address any change to the `helm` module in a separate issue after considering what the impact would be and what the desired behavior should be.

@gravesm the second example here is already part of the test suite and has been broken with this change.
Indeed we need to clarify how this should work before pushing any PR

gravesm:
I think it's fine to push this PR, but only with changes to `helm_info`. The `helm` module is a whole other mess that needs to be untangled.

Akasurde:
@RuriRyan What do you think about @gravesm's comment? 

needs_info

Issue #191, opened 2021-07-28
Add dependency update command to Helm module


WissemChb:
##### SUMMARY
I'm using the `Kubernetes.core.helm` module to deploy charts. But, if I'm not mistaken,  I found that there is no way to pull helm sub-charts when using helm dependency.

##### ISSUE TYPE
- Add the command `helm dependency update` in the module.

##### COMPONENT NAME
helm

##### ADDITIONAL INFORMATION
We add an Ansible task parameter `dependency_update: true|false(default false)` that enable/disable the helm dependency update.

```yaml
- name: Deploy Grafana chart using values files on target
  kubernetes.core.helm:
    name: test
    chart_ref: stable/grafana
    dependency_update: true #Enable dependency update
    release_namespace: monitoring
    values_files:
      - /path/to/values.yaml
```

gravesm:
@WissemChb Thanks for filing this feature request. You are correct that there is currently no way to do this using the helm module.

WissemChb:
Thank you for your response, so I'll start adding this feature.

PR #164, opened 2021-07-08
helm: add support for history-max parameter

vvanouytsel:
##### SUMMARY
The --history-max parameter allows the user to set a maximum amount of
revisions per release to be kept in the history.

By default helm keeps 10 revisions per release, which means that 10
secrets will be kept per release. If you have multiple helm releases, the amount of secrets created by Helm will quickly add up.

With supporting the 'history-max' CLI parameter via the ansible module, the user will have the posibility to determine how many revisions he wants to keep per release.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
plugins/modules/helm.py

##### ADDITIONAL INFORMATION

**Before change:**
```yaml
- name: example
  hosts: localhost
  tasks:
    - name: "Deploy helm chart before change"
      kubernetes.core.helm:
        name: "example"
        chart_ref: "/tmp/example"
        release_namespace: default
        wait_timeout: 10m
```
```
PLAY [example] ***********************************************************************************************************************************************************************************************************************

TASK [Gathering Facts] ***************************************************************************************************************************************************************************************************************
ok: [localhost]

TASK [Deploy helm chart before change] ***********************************************************************************************************************************************************************************************
changed: [localhost] => {"changed": true, "command": "/usr/bin/helm upgrade -i --reset-values example /tmp/example", "status": {"app_version": "3.0.0", "chart": "example-3.0.0", "name": "example", "namespace": "default", "revision": "1", "status": "deployed", "updated": "2021-07-08 08:38:47.591116754 +0000 UTC", "values": {}}, "stderr": "", "stderr_lines": [], "stdout": "Release \"example\" does not exist. Installing it now.\nNAME: example\nLAST DEPLOYED: Thu Jul  8 08:38:47 2021\nNAMESPACE: default\nSTATUS: deployed\nREVISION: 1\nTEST SUITE: None\nNOTES:\nDeployed!\n", "stdout_lines": ["Release \"example\" does not exist. Installing it now.", "NAME: example", "LAST DEPLOYED: Thu Jul  8 08:38:47 2021", "NAMESPACE: default", "STATUS: deployed", "REVISION: 1", "TEST SUITE: None", "NOTES:", "Deployed!"]}

PLAY RECAP ***************************************************************************************************************************************************************************************************************************
localhost                  : ok=2    changed=1    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0   
```

**After change:**
Without history_max  set:
```yaml
- name: example
  hosts: localhost
  tasks:
[...]

PR #208, opened 2021-08-17
Add helm dependency update

WissemChb:
##### SUMMARY

Execute the helm dependency update under the hood when found `dependencies` block in Chart.yaml file.
Support the execution of:

- Standalone dependency update by executing: `helm dependency update CHART` 
- Inline dependency update when specifying the helm `chart_repo_url` by adding `--dependency-update` to the `helm install` command.

##### ISSUE TYPE

- Feature Pull Request #191 

##### COMPONENT NAME

helm, helm_template
##### ADDITIONAL INFORMATION

- There is a doc generated for `history_max` option for the helm module. I think that is not generated in the previous PR #164.
- There is others changes affect the `docs/` folder when I run the `collection_prep_add_docs -p . ` command. These changes are added in the last commit  64eab40. I let you decide rather we keep the commit or remove it.

- The `--dependency-update` insertion option is tested used a local helm chart repository create via docker. So here are the tasks that test this feature.  Maybe if we create a GitHub repository for the helm chart, we can add this test code in the CI pipeline.

```paste below
# Test The update dependency with chart_repo_url
- name: "Test chart without dependencies block and chart_repo_url defined"
  block:
    - name: "Test chart without dependencies block and chart_repo_url defined"
      helm:
        binary_path: "{{ helm_binary }}"
        name: test
        chart_ref: "ingress-nginx"
        chart_repo_url: https://kubernetes.github.io/ingress-nginx
        chart_version: "{{ chart_source_version | default(omit) }}"
        namespace: "{{ helm_namespace }}"
        create_namespace: yes
      register: release

    - assert:
        that:
          - "'--dependency-update' not in release.command"
          - "'upgrade' in release.command"
        success_msg: "Command does not contains '--dependency-update' options"
        fail_msg: "Command contains '--dependency-update' options"

- name: "Test chart with dependencies block and chart_repo_url defined and replace True"
  block:
    - name: "Test chart with dependencies block and chart_repo_url defined and replace True"
      helm:
        binary_path: "{{ helm_binary }}"
        name: test1
        chart_ref: "dep_up"
        chart_repo_url: http://repo:8080/charts
        chart_version: "{{ chart_source_version | default(omit) }}"
        namespace: "{{ helm_namespace }}"
        create_namespace: yes
        replace: true
      register: release
    - debug: var=release
    - assert:
        that:
          - "'--dependency-update' in release.command"
          - "'install' in release.command"
        success_msg: "Command contains '--dependency-update' options with helm install command"
        fail_msg: "Command not contains '--dependency-update' with helm install command"

- name: "Test chart with dependencies block and chart_repo_url defined and replace False fails"
  block:
    - name: "Test chart with dependencies block and chart_repo_url defined and replace False fails"
      helm:
        binary_path: "{{ helm_binary }}"
        name: test2
        chart_ref: "dep_up"
        chart_repo_url: http://repo:8080/charts
        chart_version: "{{ chart_source_version | default(omit) }}"
        namespace: "{{ helm_namespace }}"
        create_namespace: yes
        replace: false
      register: release
      ignore_errors: true

    - assert:
        that:
          - release.failed
          - release.msg == "'--dependency-update' hasn't been supported yet with 'helm upgrade'. Please use 'helm install' instead by adding 'replace' option"
        success_msg: "Command build fail when adding  '--dependency-update' with the helm upgrade command"
```

### DIFF
--- a/docs/kubernetes.core.k8s_json_patch_module.rst
+++ b/docs/kubernetes.core.k8s_json_patch_module.rst
@@ -18,7 +18,7 @@ Version added: 2.0.0
 Synopsis
 --------
 - This module is used to apply RFC 6902 JSON patch operations only.
-- Use the :ref:`k8s <k8s_module>` module for strategic merge or JSON merge operations.
+- Use the :ref:`kubernetes.core.k8s <kubernetes.core.k8s_module>` module for strategic merge or JSON merge operations.
 - The jsonpatch library is required for check mode.
 
 
--- a/docs/kubernetes.core.k8s_json_patch_module.rst
+++ b/docs/kubernetes.core.k8s_json_patch_module.rst
@@ -18,7 +18,7 @@ Version added: 2.0.0
 Synopsis
 --------
 - This module is used to apply RFC 6902 JSON patch operations only.
-- Use the :ref:`kubernetes.core.k8s <kubernetes.core.k8s_module>` module for strategic merge or JSON merge operations.
+- Use the :ref:`k8s <k8s_module>` module for strategic merge or JSON merge operations.
 - The jsonpatch library is required for check mode.
 
 
--- a/docs/kubernetes.core.k8s_log_module.rst
+++ b/docs/kubernetes.core.k8s_log_module.rst
@@ -387,6 +387,22 @@ Parameters
             <tr>
                 <td colspan="2">
                     <div class="ansibleOptionAnchor" id="parameter-"></div>
+                    <b>since_seconds</b>
+                    <a class="ansibleOptionLink" href="#parameter-" title="Permalink to this option"></a>
+                    <div style="font-size: small">
+                        <span style="color: purple">string</span>
+                    </div>
+                    <div style="font-style: italic; font-size: small; color: darkgreen">added in 2.2.0</div>
+                </td>
+                <td>
+                </td>
+                <td>
+                        <div>A relative time in seconds before the current time from which to show logs.</div>
+                </td>
+            </tr>
+            <tr>
+                <td colspan="2">
+                    <div class="ansibleOptionAnchor" id="parameter-"></div>
                     <b>username</b>
                     <a class="ansibleOptionLink" href="#parameter-" title="Permalink to this option"></a>
                     <div style="font-size: small">
@@ -458,6 +474,7 @@ Examples
         kind: Deployment
         namespace: testing
         name: example
+        since_seconds: "4000"
       register: log
 
     # This will get the log from a single Pod managed by this DeploymentConfig
--- a/docs/kubernetes.core.k8s_log_module.rst
+++ b/docs/kubernetes.core.k8s_log_module.rst
@@ -387,22 +387,6 @@ Parameters
             <tr>
                 <td colspan="2">
                     <div class="ansibleOptionAnchor" id="parameter-"></div>
-                    <b>since_seconds</b>
-                    <a class="ansibleOptionLink" href="#parameter-" title="Permalink to this option"></a>
-                    <div style="font-size: small">
-                        <span style="color: purple">string</span>
-                    </div>
-                    <div style="font-style: italic; font-size: small; color: darkgreen">added in 2.2.0</div>
-                </td>
-                <td>
-                </td>
-                <td>
-                        <div>A relative time in seconds before the current time from which to show logs.</div>
-                </td>
-            </tr>
-            <tr>
-                <td colspan="2">
-                    <div class="ansibleOptionAnchor" id="parameter-"></div>
                     <b>username</b>
                     <a class="ansibleOptionLink" href="#parameter-" title="Permalink to this option"></a>
                     <div style="font-size: small">
@@ -474,7 +458,6 @@ Examples
         kind: Deployment
         namespace: testing
         name: example
-        since_seconds: "4000"
       register: log
 
     # This will get the log from a single Pod managed by this DeploymentConfig
--- a/docs/kubernetes.core.k8s_lookup.rst
+++ b/docs/kubernetes.core.k8s_lookup.rst
@@ -381,12 +381,6 @@ Parameters
     <br/>
 
 
-Notes
------
-
-.. note::
-   - While querying, please use ``query`` or ``lookup`` format with ``wantlist=True`` to provide an easier and more consistent interface. For more details, see https://docs.ansible.com/ansible/latest/plugins/lookup.html#forcing-lookups-to-return-lists-query-and-wantlist-true.
-
 
 
 Examples
@@ -396,23 +390,23 @@ Examples
 
     - name: Fetch a list of namespaces
       set_fact:
-        projects: "{{ query('kubernetes.core.k8s', api_version='v1', kind='Namespace') }}"
+        projects: "{{ lookup('kubernetes.core.k8s', api_version='v1', kind='Namespace') }}"
 
     - name: Fetch all deployments
       set_fact:
-        deployments: "{{ query('kubernetes.core.k8s', kind='Deployment') }}"
+        deployments: "{{ lookup('kubernetes.core.k8s', kind='Deployment') }}"
 
     - name: Fetch all deployments in a namespace
       set_fact:
-        deployments: "{{ query('kubernetes.core.k8s', kind='Deployment', namespace='testing') }}"
+        deployments: "{{ lookup('kubernetes.core.k8s', kind='Deployment', namespace='testing') }}"
 
     - name: Fetch a specific deployment by name
       set_fact:
-        deployments: "{{ query('kubernetes.core.k8s', kind='Deployment', namespace='testing', resource_name='elastic') }}"
+        deployments: "{{ lookup('kubernetes.core.k8s', kind='Deployment', namespace='testing', resource_name='elastic') }}"
 
     - name: Fetch with label selector
       set_fact:
-        service: "{{ query('kubernetes.core.k8s', kind='Service', label_selector='app=galaxy') }}"
+        service: "{{ lookup('kubernetes.core.k8s', kind='Service', label_selector='app=galaxy') }}"
 
     # Use parameters from a YAML config
 
@@ -422,11 +416,11 @@ Examples
 
     - name: Using the config (loaded from a file in prior task), fetch the latest version of the object
       set_fact:
-        service: "{{ query('kubernetes.core.k8s', resource_definition=config) }}"
+        service: "{{ lookup('kubernetes.core.k8s', resource_definition=config) }}"
 
     - name: Use a config from the local filesystem
       set_fact:
-        service: "{{ query('kubernetes.core.k8s', src='service.yml') }}"
+        service: "{{ lookup('kubernetes.core.k8s', src='service.yml') }}"
 
 
 
--- a/docs/kubernetes.core.k8s_lookup.rst
+++ b/docs/kubernetes.core.k8s_lookup.rst
@@ -381,6 +381,12 @@ Parameters
     <br/>
 
 
+Notes
+-----
+
+.. note::
+   - While querying, please use ``query`` or ``lookup`` format with ``wantlist=True`` to provide an easier and more consistent interface. For more details, see https://docs.ansible.com/ansible/latest/plugins/lookup.html#forcing-lookups-to-return-lists-query-and-wantlist-true.
+
 
 
 Examples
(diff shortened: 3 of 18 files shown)


gravesm:
@WissemChb thanks for the PR! This is looking like a good start. I think the better way to do this is to add a `dependency_update` parameter as you initially suggested in your issue, though. This PR forces the update which would change the existing behavior and also not allow any control over whether or not to do the update.

With regards to the docs changes, you should restore all the deleted .rst files. The doc generation command is pretty rough and ideally shouldn't be deleting those. The other changes you can keep. We always generate the docs right before a release but it's fine to go ahead and run it now, too.

WissemChb:
> @WissemChb thanks for the PR! This is looking like a good start. I think the better way to do this is to add a `dependency_update` parameter as you initially suggested in your issue, though. This PR forces the update which would change the existing behavior and also not allow any control over whether or not to do the update.
> 
> With regards to the docs changes, you should restore all the deleted .rst files. The doc generation command is pretty rough and ideally shouldn't be deleting those. The other changes you can keep. We always generate the docs right before a release but it's fine to go ahead and run it now, too.

Hello @gravesm, Thank you for your reply. I'll take into consideration your suggestions. Initially, I considered adding the `dependency_update` options but when I go through the dev process I decide to simplify the module params and options. But you are right it's better to add more control over the update process.

molecule/default/roles/helm/tasks/test_up_dep.yml
```diff
@@ -0,0 +1,175 @@
...
+      register: stat_result
+
+    - assert:
+        that:
+          - stat_result.stat.exists
+        success_msg: "subchart exist in the chart directory"
+        fail_msg: "subchart not exist in the charts directory"
+
+# Clean
+- name: "Clean"
```
gravesm, line 79: This needs to be in an `always` block. The best way to do this will be to put everything in this file except this block into a single block and then put this part into an `always` block at the bottom. Look at tests_helm_diff.yml for an example.

molecule/default/roles/helm/tasks/test_up_dep.yml
```diff
@@ -0,0 +1,175 @@
...
+    - name: Test Helm subchart not pulled when dependency_update false
+      helm_template:
+        binary_path: "{{ helm_binary }}"
+        chart_ref: "/tmp/test-chart"
+        chart_version: "{{ chart_source_version | default(omit) }}"
+        dependency_update: false
+        output_dir: "/tmp"
+      register: result
+
+    - name: "Check if the subchart exist in chart"
```
gravesm, line 162: This should be testing that `/tmp/test-chart/charts` does not exist.

plugins/modules/helm.py
```diff
@@ -46,6 +46,16 @@
...
+  dependency_update:
+    description:
+      - Run standelone C(helm dependency update CHART) before the operation.
+      - Run inline C(--dependency-update) with C(helm install) command. This feature is not supported yet with the C(helm upgrade) command.
+      - So we should consider to use I(dependency_update) options with I(replace) option enabled when specifying I(chart_repo_url).
+      - The I(dependency_update) option require the add of C(dependencies) block in C(Chart.yaml/requirements.yaml) file.
+      - For more information please visit U(https://helm.sh/docs/helm/helm_dependency/)
+    default: false
+    type: bool
+    aliases: [ dep_up ]
```
gravesm, line 58:
```suggestion
    aliases: [ dep_up ]
    version_added: "2.2.0"
```

plugins/modules/helm_template.py
```diff
@@ -45,6 +45,14 @@
...
     required: false
     type: str
+  dependency_update:
+    description:
+      - Run helm dependency update before the operation.
+      - The I(dependency_update) option require the add of C(dependencies) block in C(Chart.yaml/requirements.yaml) file.
+      - For more information please visit U(https://helm.sh/docs/helm/helm_dependency/)
+    default: false
+    type: bool
+    aliases: [ dep_up ]
```
gravesm, line 55:
```suggestion
    aliases: [ dep_up ]
    version_added: "2.2.0"
```

gravesm (changes requested):
Thanks for your work on this, and for adding tests! In addition to inline comments we'll need a changelog fragment: https://docs.ansible.com/ansible/latest/community/development_process.html#changelogs

plugins/modules/helm.py
```diff
@@ -612,17 +634,38 @@ def main():
...
 
         # Fetch chart info to have real version and real name for chart_ref from archive, folder or url
         chart_info = fetch_chart_info(module, helm_cmd, chart_ref)
 
+        if dependency_update:
+            if not chart_info.get('dependencies'):
+                msg_fail = (f"No subchart will be pulled for {chart_ref}. "
+                            "Please make sure to add dependencies block in Chart.yaml or requirements.yaml. "
+                            "For more information please visite https://helm.sh/docs/helm/helm_dependency/")
+                module.fail_json(msg=msg_fail)
```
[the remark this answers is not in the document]
WissemChb: I hesitated between adding only a warning message or fail the execution. And I think I chose the wrong option. So I'll add a warning message instead.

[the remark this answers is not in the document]
WissemChb: I hesitated between a warn message or a fail message and I think that I chose the wrong option. So l'il replace this block with a warn message.

plugins/modules/helm.py
```diff
@@ -612,17 +634,38 @@ def main():
...
+            # we can get rid of this method and use only '--dependency-update'
+            # option. Please see https://github.com/helm/helm/pull/8810
+            if not chart_repo_url:
+                run_dep_update(module, helm_cmd_common, chart_ref)
+            else:
+                module.warn("This is a not stable feature with 'chart_repo_url'. Please consider to use dependency update with on-disk charts")
+                if not replace:
+                    msg_fail = ("'--dependency-update' hasn't been supported yet with 'helm upgrade'. "
+                                "Please use 'helm install' instead by adding 'replace' option")
+                    module.fail_json(msg=msg_fail)
```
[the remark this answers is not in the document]
WissemChb: This part is becoming very confusing for me. for 2 reasons:
1. What I found in Ansible doc is that if I make `dependency_update` and `replace` mutually exclusive I can't enable both of them and that is not the case because when the `chart_ref_url` is defined or `chart_ref` is an URL I need to enable `replace` option when `dependency_update` is enabled to use `helm install` instead of `helm update`(not support `--dependency-update` option. I'm sorry, this is my first ansible module creation maybe I need more details about your suggestion.
2. After doing a deep search I realised that we haven't the need to do `dependency update` with remote charts because to package them we need to pull the subchart by running `helm package --dependency-update dep-up` So, if I'm not mistaken, all remote package that depends on other charts have their subchart pulled already within them.

plugins/modules/helm.py
```diff
@@ -612,17 +634,38 @@ def main():
...
+                            "Please make sure to add dependencies block in Chart.yaml or requirements.yaml. "
+                            "For more information please visite https://helm.sh/docs/helm/helm_dependency/")
+                module.fail_json(msg=msg_fail)
+
+            # Can't use '--dependency-update' with 'helm upgrade' that is the
+            # default chart install method, so if chart_repo_url is defined
+            # we can't use the dependency update command. But, in the near future
+            # we can get rid of this method and use only '--dependency-update'
+            # option. Please see https://github.com/helm/helm/pull/8810
+            if not chart_repo_url:
```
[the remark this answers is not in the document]
WissemChb: related to the previous reply

docs/kubernetes.core.k8s_json_patch_module.rst
```diff
@@ -18,7 +18,7 @@ Version added: 2.0.0
 Synopsis
 --------
 - This module is used to apply RFC 6902 JSON patch operations only.
-- Use the :ref:`k8s <k8s_module>` module for strategic merge or JSON merge operations.
+- Use the :ref:`kubernetes.core.k8s <kubernetes.core.k8s_module>` module for strategic merge or JSON merge operations.
```
abikouo, line 21: This has nothing to do with this PR, could you please revert this file ? 
Thanks

```suggestion
- Use the :ref:`k8s <k8s_module>` module for strategic merge or JSON merge operations.
```

docs/kubernetes.core.k8s_log_module.rst
```diff
@@ -383,6 +383,22 @@ Parameters
...
+                        <span style="color: purple">string</span>
+                    </div>
+                    <div style="font-style: italic; font-size: small; color: darkgreen">added in 2.2.0</div>
+                </td>
+                <td>
+                </td>
+                <td>
+                        <div>A relative time in seconds before the current time from which to show logs.</div>
+                </td>
+            </tr>
```
abikouo, line 401: please revert this modification. the doc for this module should be modified

docs/kubernetes.core.k8s_lookup.rst
```diff
@@ -381,6 +381,12 @@ Parameters
     <br/>
 
 
+Notes
+-----
+
+.. note::
+   - While querying, please use ``query`` or ``lookup`` format with ``wantlist=True`` to provide an easier and more consistent interface. For more details, see https://docs.ansible.com/ansible/latest/plugins/lookup.html#forcing-lookups-to-return-lists-query-and-wantlist-true.
+
```
abikouo, line 389: same for this file, should not be modified

plugins/modules/helm.py
```diff
@@ -46,6 +46,17 @@
       - Chart version to install. If this is not specified, the latest version is installed.
     required: false
     type: str
+  dependency_update:
+    description:
+      - Run standelone C(helm dependency update CHART) before the operation.
```
abikouo, line 51:
```suggestion
      - Run standalone C(helm dependency update CHART) before the operation.
```

plugins/modules/helm.py
```diff
@@ -612,17 +635,39 @@ def main():
         if chart_version is not None:
             helm_cmd += " --version=" + chart_version
 
-        if chart_repo_url is not None:
+        if chart_repo_url:
```
abikouo, line 638: why this change ? 
In some case it is better to check ``None`` value instead of implicit conversion of str to bool

plugins/modules/helm.py
```diff
@@ -612,17 +635,39 @@ def main():
...
+                            "Please consider add dependencies block or disable dependency_update to remove this warning.")
+
+            # Can't use '--dependency-update' with 'helm upgrade' that is the
+            # default chart install method, so if chart_repo_url is defined
+            # we can't use the dependency update command. But, in the near future
+            # we can get rid of this method and use only '--dependency-update'
+            # option. Please see https://github.com/helm/helm/pull/8810
+            if not chart_repo_url:
+                run_dep_update(module, helm_cmd_common, chart_ref)
+
```
abikouo, line 656:
```suggestion
      else:
            # Can't use '--dependency-update' with 'helm upgrade' that is the
            # default chart install method, so if chart_repo_url is defined
            # we can't use the dependency update command. But, in the near future
            # we can get rid of this method and use only '--dependency-update'
            # option. Please see https://github.com/helm/helm/pull/8810
            if not chart_repo_url:
                run_dep_update(module, helm_cmd_common, chart_ref)

```

plugins/modules/helm.py
```diff
@@ -612,17 +635,39 @@ def main():
...
             helm_cmd += " --repo=" + chart_repo_url
 
         # Fetch chart info to have real version and real name for chart_ref from archive, folder or url
         chart_info = fetch_chart_info(module, helm_cmd, chart_ref)
 
+        if dependency_update:
+            if not chart_info.get('dependencies'):
+                module.warn("There is no dependencies block defined in Chart.yaml. Dependency update will not be performed. "
+                            "Please consider add dependencies block or disable dependency_update to remove this warning.")
+
```
abikouo, line 648: why not putting else here to avoid running ``helm dependency update`` ?

plugins/modules/helm.py
```diff
@@ -612,17 +635,39 @@ def main():
...
+            # Can't use '--dependency-update' with 'helm upgrade' that is the
+            # default chart install method, so if chart_repo_url is defined
+            # we can't use the dependency update command. But, in the near future
+            # we can get rid of this method and use only '--dependency-update'
+            # option. Please see https://github.com/helm/helm/pull/8810
+            if not chart_repo_url:
+                run_dep_update(module, helm_cmd_common, chart_ref)
```
abikouo, line 655: please consider returning at least the result code ``rc`` which will validate that the dependency update has succeed or return error message if not

plugins/modules/helm.py
```diff
@@ -352,6 +363,14 @@ def run_repo_update(module, command):
     rc, out, err = run_helm(module, repo_update_command)
 
 
+def run_dep_update(module, command, chart_ref):
+    """
+    Run dependency update
+    """
+    dep_update = command + " dependency update " + chart_ref
```
abikouo, line 370: this will fail when the ``chart_ref`` contains an URL, you need to check somewhere that it is not the case

plugins/modules/helm.py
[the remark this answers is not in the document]
abikouo: agree, ``chart_ref`` should be validated as non URL

WissemChb:
@gravesm and @abikouo, I added what you suggested. Can you check, please ? Thank you very much for your the time to make the review.
> one of the three sanity tests fails in the Zuul pipeline can you please check and tell me what I missed because the test worked fine in local for me.

abikouo:
recheck

abikouo (changes requested):

joestump:
Got bit by this today. Anything I can do to help?

gravesm:
Sorry, this one appears to have fallen through the cracks. I've gone ahead and rebased, and fixed a few things to account for recent changes to our CI.

gravesm:
recheck

abikouo (approved):
LGTM!!

gravesm (approved):

PR #218, opened 2021-08-31
Disable turbo mode for the validate tests

gravesm:
Depends-On: ansible/ansible-zuul-jobs#1074

Switching virtualenvs in the test suite does not work so well with turbo
mode enabled because the module may not be executed by the virtualenv
that is specified for that task.

### DIFF
--- a/molecule/default/tasks/validate.yml
+++ b/molecule/default/tasks/validate.yml
@@ -236,3 +236,5 @@
 
   vars:
     validate_namespace: validate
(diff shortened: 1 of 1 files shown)


abikouo (approved):
LGTM

goneri (approved):

Akasurde:
@gravesm Thanks for the PR. @goneri Thanks for reviewing.

PR #220, opened 2021-09-01
add host and api_key properties to helm_repository

6485im43:
##### SUMMARY
add posibility to manage clusters via `host` and `api_token` instead of using kubeconfig in helm_repository.py. Code and properties are the same as already implemented in helm.py.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
extend helm_repository

### DIFF
--- a/docs/kubernetes.core.helm_repository_module.rst
+++ b/docs/kubernetes.core.helm_repository_module.rst
@@ -43,6 +43,22 @@ Parameters
             <tr>
                 <td colspan="1">
                     <div class="ansibleOptionAnchor" id="parameter-"></div>
+                    <b>api_key</b>
+                    <a class="ansibleOptionLink" href="#parameter-" title="Permalink to this option"></a>
+                    <div style="font-size: small">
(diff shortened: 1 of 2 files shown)


gravesm (changes requested):
@6485im43 Thank you for the PR! The docs are autogenerated from the module doc string so you should revert your existing doc changes. In addition to adding the parameters you will need to update the doc string on the module. You can just include the helm common options like it was done in the helm module here: https://github.com/ansible-collections/kubernetes.core/blob/8805a20c63b4f1774bcd9c75d352f5b0899fdac4/plugins/modules/helm.py#L149-L150

Could you also provide a changelog fragment? https://docs.ansible.com/ansible/latest/community/development_process.html#community-changelogs

Akasurde:
@6485im43 Are you still working on this? Thanks.

Akasurde:
@6485im43 Thanks for the contribution. 

superseded by https://github.com/ansible-collections/kubernetes.core/pull/354

PR #233, opened 2021-09-13
Add deprecation notice to k8s_exec

gravesm:
##### ISSUE TYPE

- Bugfix Pull Request

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/plugins/modules/k8s_exec.py
+++ b/plugins/modules/k8s_exec.py
@@ -183,7 +183,7 @@ def execute_module(module, k8s_ansible_mixin):
     else:
         rc = int(err['details']['causes'][0]['message'])
 
-    module.deprecate("The 'return_code' return key is deprecated. Please use 'rc' instead.", date="2022-10-01", collection_name="kubernetes.core")
+    module.deprecate("The 'return_code' return key is deprecated. Please use 'rc' instead.", version="4.0.0", collection_name="kubernetes.core")
     module.exit_json(
         # Some command might change environment, but ultimately failing at end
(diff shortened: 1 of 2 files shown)


gravesm:
recheck

gravesm:
recheck

plugins/modules/k8s_exec.py
```diff
@@ -183,6 +183,7 @@ def execute_module(module, k8s_ansible_mixin):
     else:
         rc = int(err['details']['causes'][0]['message'])
 
+    module.deprecate("The 'return_code' return key is deprecated. Please use 'rc' instead.", version="4.0.0", collection_name="kubernetes.core")
```
abikouo, line 186: how to do manage redirection for this deprecation ? if someone is using ``community.kubernetes.k8s_exec`` for instance

gravesm: What do you mean? The deprecation warning will be printed the same if you are using community.kubernetes. Sure, it will say "kubernetes.core", but you will also have gotten a deprecation warning that community.kubernetes is deprecated. I don't think we want to try and do anything to have it print community.kubernetes because that would be a ton of work for a collection we no longer are supporting.

alinabuzachis (approved):

plugins/modules/k8s_exec.py
abikouo: agree.
I was just trying to understand.
Thanks

abikouo (approved):

PR #236, opened 2021-09-16
add diff output when using helmdiff

joschi36:
I have tried it before here: https://github.com/ansible-collections/community.kubernetes/pull/363
copied from my other PR:

> ##### SUMMARY
> In the previous PR https://github.com/ansible-collections/community.kubernetes/pull/355 @Akasurde has implemented optional helm check by the helm plugin diff.
> I was unsure why we would not use this for --diff in Ansible. This would be very useful.
> I have added a gist with a simple module I build which is actually using it.
> ~~Also, I think this PR needs some more love, but first I wanted to gather a little bit of feedback from the original creator @Akasurde and @LucasBoisserie~~
> 
> ##### ISSUE TYPE
> 
> - Feature Pull Request
> 
> ##### COMPONENT NAME
> 
> Helm
> 
> ##### ADDITIONAL INFORMATION
> - 
>

### DIFF
--- a/changelogs/fragments/236-add-diff-output-to-helm-module.yaml
+++ b/changelogs/fragments/236-add-diff-output-to-helm-module.yaml
@@ -1,0 +1,2 @@
+bugfixes:
+  - helm - If using module with helmdiff, the module now outputs diff (https://github.com/ansible-collections/kubernetes.core/pull/236).
--- a/plugins/modules/helm.py
+++ b/plugins/modules/helm.py
@@ -565,6 +565,7 @@ def main():
         module.fail_json(msg=missing_required_lib("yaml"), exception=IMP_YAML_ERR)
 
     changed = False
+    diff = ""
 
     bin_path = module.params.get('binary_path')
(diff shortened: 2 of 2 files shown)


joschi36:
recheck

Akasurde:
recheck

gravesm (changes requested):
Thanks, @joschi36, this looks good. The only thing that needs to be addressed is that we should specifically check if the user has requested a diff. You can use the  `module._diff` boolean variable to check this. If it's false, just set the diff back to an empty string.

Akasurde:
@joschi36 Could you please rebase this PR and ping me? Thanks

joschi36:
@Akasurde Sorry that there was no movement in the last few months.

I think @jfrabaute's Pull Request https://github.com/ansible-collections/kubernetes.core/pull/290 supersedes this. Can you confirm?

Akasurde:
@joschi36 Thanks for the confirmation. It does look like #290 supersedes this PR. I will go ahead and close this.

Thanks for the contribution.

Issue #35, opened 2020-08-06
support generateName or utils like that


alaypatel07:
##### SUMMARY

In order to make sure that resources created by k8s module has unique
name, there is some work involded to be done in ansible. kubernetes
api-server has a `metadata.generateName` field which can be used
to create a name that is within the limits of max chars allowed for the 
resource. It would be great if the k8s module allows creation of 
resources with generateName so the user does not have to take
care of limits of `metadata.Name`

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

k8s

##### ADDITIONAL INFORMATION

```yaml
cat <<EOF | kubectl create -f -
apiVersion: v1
kind: Pod
metadata:
  generateName: utilitypod-
  namespace: blah-dev
  labels:
    purpose: utility-pod
spec:
  containers:
    - name: utilitypod
      image: blahblah/utilitypod:latest
      command: [ "/bin/bash", "-c", "--" ]
      args: [ "while true; do sleep 28800; done;" ]
      env:
        - name: KUBERNETES_SERVICE_HOST
          value: "api.dev.blah.internal"
        - name: KUBERNETES_SERVICE_PORT
          value: "443"
EOF

```

Taken from this stackoverflow thread https://stackoverflow.com/questions/48023475/add-random-string-on-kubernetes-pod-deployment-name

abikouo:
@alaypatel07 @FWiesner @jkupferer collection now support ``generateName`` added in #238 
Feel free to test it and provide feedback.
Thanks

FWiesner:
@abikouo thanks!

PR #238, opened 2021-09-17
k8s - allow resource definition using generateName

abikouo:
##### SUMMARY

#35 

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

k8s
##### ADDITIONAL INFORMATION

```paste below
- name: create pod using generateName
  k8s:
    namespace: test
    generate_name: pod-
    definition:
       kind: Pod
       spec:
          containers:
          - name: py
            image: python:3.7-alpine

- name: create pod using generateName
  k8s:
    namespace: test
    definition:
       kind: Pod
       metadata:
          generateName: pod-
       spec:
          containers:
          - name: py
            image: python:3.7-alpine

```


gravesm (changes requested):
There are a few problems that will need to be sorted out. In general, we'll need to go through very carefully and make sure that in every instance that we are accessing, or trying to access, the `name` that we are accounting for the changes here. Some examples of things that are not working, all for different reasons:

```
k8s:
  kind: Namespace
  generate_name: ns-
```
```
k8s:
  kind: ConfigMap
  namespace: whatever
  generate_name: cmap-
  append_hash: yes
```
```
k8s:
  definition:
    apiVersion: v1
    kind: Pod
    metadata:
      namespace: whatever
      name: pod1
    spec:
      containers:
        - image: adslfkjadslfkjadslkfjsadf
          name: non-existent-container

k8s:
  definition:
    apiVersion: v1
    kind: Pod
    metadata:
      namespace: whatever
    spec:
      containers:
        - image: busybox
          name: c1
  generate_name: pod-
  wait: yes
```
In the latter case, because there is no name, we end up waiting on every pod in that namespace. This wait will fail because the first pod has a non-existent image name.

abikouo:
> There are a few problems that will need to be sorted out. In general, we'll need to go through very carefully and make sure that in every instance that we are accessing, or trying to access, the `name` that we are accounting for the changes here. Some examples of things that are not working, all for different reasons:
> 
> ```
> k8s:
>   kind: Namespace
>   generate_name: ns-
> ```
> 
> ```
> k8s:
>   kind: ConfigMap
>   namespace: whatever
>   generate_name: cmap-
>   append_hash: yes
> ```
> 
> ```
> k8s:
>   definition:
>     apiVersion: v1
>     kind: Pod
>     metadata:
>       namespace: whatever
>       name: pod1
>     spec:
>       containers:
>         - image: adslfkjadslfkjadslkfjsadf
>           name: non-existent-container
> 
> k8s:
>   definition:
>     apiVersion: v1
>     kind: Pod
>     metadata:
>       namespace: whatever
>     spec:
>       containers:
>         - image: busybox
>           name: c1
>   generate_name: pod-
>   wait: yes
> ```
> 
> In the latter case, because there is no name, we end up waiting on every pod in that namespace. This wait will fail because the first pod has a non-existent image name.

ok for the latter case, I see the issue with the wait mechanisms, I need to check for other scenarios
Thanks @gravesm

abikouo:
@gravesm regarding the latter case you just mentioned in your comment above, it is not failing the resource name is updated before calling the wait method,
https://github.com/ansible-collections/kubernetes.core/blob/2ffe81353549e1999571d12c96e86cea6ef40276/plugins/module_utils/common.py#L824-L825
I have added test case to validate that
https://github.com/ansible-collections/kubernetes.core/blob/2ffe81353549e1999571d12c96e86cea6ef40276/molecule/default/tasks/generate_name.yml#L149-L180

For others scenarios, code has been fixed and test updated accordingly

molecule/default/tasks/generate_name.yml
```diff
@@ -0,0 +1,188 @@
...
+        apiVersion: v1
+        kind: Pod
+        metadata:
+          name: pod1
+        spec:
+          containers:
+          - image: adslfkjadslfkjadslkfjsadf
+            name: non-existent-container-image
+
+  - name: Create second Pod using wait (it should not wait for the first container)
```
gravesm, line 162: The important bit for this one is that the pod definition should *not* include a `name`, but you should use the `generate_name` param, instead. This does appear to be passing locally for me now with the new changes, but we should update the test.

abikouo: fixed.
Thanks for the feedback

abikouo:
@Akasurde could you please take a look at this one? 
Thanks

gravesm (approved):

PR #239, opened 2021-09-20
common/_wait_for: ensure label_selectors is optional

goneri:
Depends-On: https://github.com/ansible/ansible-zuul-jobs/pull/1125

The label_selectors is a new parameter for _wait_for that was
introduced in https://github.com/ansible-collections/kubernetes.core/pull/158.

The value is new and it can be set to None to make it optional. It should
not be mandatory a non optional parameter.

### DIFF
--- a/changelogs/fragments/_wait_for_label_selector_optional.yaml
+++ b/changelogs/fragments/_wait_for_label_selector_optional.yaml
@@ -1,0 +1,3 @@
+---
+bugfixes:
+- common - Ensure the label_selectors parameter of _wait_for method is optional.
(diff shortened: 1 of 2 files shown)


abikouo (approved):
LGTM

goneri:
The CI failure is suspicious, I just opened https://github.com/ansible-collections/kubernetes.core/issues/241 to trace it.

goneri:
recheck

alinabuzachis (approved):

gravesm:
recheck

PR #245, opened 2021-09-22
Add support for dry run

gravesm:
##### SUMMARY

Kubernetes server-side dry run will be used when the kubernetes client
version is >=18.20.0. For older versions of the client, the existing
client side speculative change implementation will be used.

The effect of this change should be mostly transparent to the end user
and is reflected in the fact the tests have not changed but should still
pass. With this change, there are a few edge cases that will be
improved. One example of these edge cases is to use check mode on an
existing Service resource. With dry run this will correctly report no
changes, while the older client side implementation will erroneously
report changes to the port spec.

##### ISSUE TYPE

- Feature Pull Request

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/plugins/module_utils/common.py
+++ b/plugins/module_utils/common.py
@@ -229,6 +229,7 @@ class K8sAnsibleMixin(object):
             module.fail_json(msg=missing_required_lib('kubernetes'), exception=K8S_IMP_ERR,
                              error=to_native(k8s_import_exception))
         self.kubernetes_version = kubernetes.__version__
+        self.supports_dry_run = LooseVersion(self.kubernetes_version) >= LooseVersion("18.20.0")
 
         if pyyaml_required and not HAS_YAML:
             module.fail_json(msg=missing_required_lib("PyYAML"), exception=YAML_IMP_ERR)
@@ -668,7 +669,9 @@ class K8sAnsibleMixin(object):
             else:
                 # Delete the object
                 result['changed'] = True
-                if not self.check_mode:
+                if self.check_mode and not self.supports_dry_run:
+                    return result
+                else:
                     if delete_options:
                         body = {
                             'apiVersion': 'v1',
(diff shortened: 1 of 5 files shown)


plugins/module_utils/common.py
```diff
@@ -668,14 +669,18 @@ def _empty_resource_list():
             else:
                 # Delete the object
                 result['changed'] = True
-                if not self.check_mode:
+                if self.check_mode and not self.supports_dry_run:
```
goneri, line 672: Instead of doing these checks on two values, can we instead wrap `check_mode` in a setter (e.g: `enable_check_mode()`), and check the `supports_dry_run value` inside the setter. This way, check_mode can be true only if supports_dry_run is true. And we do the check at one single place in the code.

gravesm: I may not be fully understanding what you are suggesting, but I don't think this would be possible. The problem is check_mode can be true even if the library doesn't support dry run. In that case, we use the older, slightly buggy client side dry run. So there are really three states a module could be in: 1. no check mode 2. check mode with dry run 3. check mode without dry run. Different actions will be taken for all three states.

I agree this is ugly. It's a perfect example of one of the problems I've highlighted in this collection as why it needs refactoring. This is a change that ideally happens in one place. Until we can drop support for the older kubernetes client I think it's always going to be a bit messy, but once we can solely rely on dry run the best way to handle this is to probably be able to put the client into something like read only mode. This would happen once during module setup and we wouldn't need to have any check mode logic in the rest of the code.

goneri: Thanks for taking the time to explain the problem.

alinabuzachis (approved):

abikouo:
@gravesm  is this not relevant for others modules modifying resources ? ``k8s_drain``, ``k8s_service``, ``k8s_rollback`` and ``k8s_scale``.
``k8s_drain`` does not support the check_mode yet.

gravesm:
@abikouo Yes, this can and eventually should be used for all of those modules, but none of them currently support check mode. That's work that should probably be done in a separate PR. In fact, k8s_rollback and k8s_scale actually claim to support check mode, but don't. I have already created issues for both of those.

abikouo (approved):

PR #253, opened 2021-10-04
Don't wait on *List resources for info module

gravesm:
##### SUMMARY

We can't use the same wait logic on *List resources because they lack
the same metadata that other resources have. We should ensure that we
are waiting on the items in the list, but not the list itself. Waiting
on the list itself results in unexpected behavior.

This fixes the waiting logic when waiting on a list to wait until the
list being queried contains one or more items, or the wait timeout has
been reached. Each item in the list can then be waited on with the usual
wait logic.

##### ISSUE TYPE

- Bugfix Pull Request

##### ADDITIONAL INFORMATION

```paste below

```


gravesm:
recheck

gravesm:
recheck

PR #254, opened 2021-10-05
k8s_cp turbo mode compliance

### LINKED ISSUE #237: k8s_cp does not delay module load
##### SUMMARY

update: k8s_cp does not use the turbo mode at all.

Before `k8s_cp`'s main() is called, all the resources have to be loaded. This has an negative impact of the performance boost that we can get with turbo mode. Turbo mode works in two steps:

- regular: load the module, evaluate the parameters and send the module to the background daemon.
- in the background daemon: load the module and run the main() function.

The second steps should be fast because these dependencies are likely to be already in memory. We don't need most of k8s_cp.py during the first step. This content should be moved in a module_utils file that we lazy-load.

##### ISSUE TYPE
- Bug Report

abikouo:
##### SUMMARY

closes #237 

##### ISSUE TYPE

- Feature Pull Request
##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/plugins/modules/k8s_cp.py
+++ b/plugins/modules/k8s_cp.py
@@ -138,424 +138,43 @@ result:
 """
 
 import copy
-import os
-from tempfile import TemporaryFile, NamedTemporaryFile
-from select import select
-from abc import ABCMeta, abstractmethod
-import tarfile
 
-# from ansible_collections.kubernetes.core.plugins.module_utils.ansiblemodule import AnsibleModule
-from ansible.module_utils.basic import AnsibleModule
-from ansible.module_utils._text import to_native
-from ansible_collections.kubernetes.core.plugins.module_utils.common import (
-    K8sAnsibleMixin,
-    get_api_client,
+from ansible_collections.kubernetes.core.plugins.module_utils.ansiblemodule import (
+    AnsibleModule,
 )
 from ansible_collections.kubernetes.core.plugins.module_utils.args_common import (
     AUTH_ARG_SPEC,
 )
-
-try:
-    from kubernetes.client.api import core_v1_api
-    from kubernetes.stream import stream
-    from kubernetes.stream.ws_client import (
-        STDOUT_CHANNEL,
-        STDERR_CHANNEL,
-        ERROR_CHANNEL,
-        ABNF,
-    )
-except ImportError:
-    pass
-
-try:
-    import yaml
-except ImportError:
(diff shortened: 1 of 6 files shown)


plugins/modules/k8s_cp.py
```diff
@@ -573,32 +192,17 @@ def execute_module(module):
...
-        k8s_copy = load_class.get(state)(module, k8s_ansible_mixin.client)
-        k8s_copy.run()
-    except Exception as e:
-        module.fail_json("Failed to copy object due to: {0}".format(to_native(e)))
+    state = module.params.get("state")
+    if state == "to_pod":
+        k8s_copy = K8SCopyToPod(module, k8s_ansible_mixin.client)
+    else:
+        k8s_copy = K8SCopyFromPod(module, k8s_ansible_mixin.client)
+    k8s_copy.run()
```
gravesm, line 200: Should we wrap this `run()` call in a try/except block?

alinabuzachis (commented):
Am I wrong or the check_mode logic is not handled here (even if the module has supports_check_mode=True)?

abikouo:
> Am I wrong or the check_mode logic is not handled here (even if the module has supports_check_mode=True)?

check_mode is not supported by the module even if it states that, but this should be addressed in another PR
This PR is resolving turbo mode compliance only

alinabuzachis:
> > Am I wrong or the check_mode logic is not handled here (even if the module has supports_check_mode=True)?
> 
> check_mode is not supported by the module even if it states that, but this should be addressed in another PR This PR is resolving turbo mode compliance only

So, I was just wonder if it would be better to have it set to False if it's not supported. @gravesm

gravesm:
> check_mode is not supported by the module even if it states that, but this should be addressed in another PR
This PR is resolving turbo mode compliance only

That's fine. Please, create an issue for it.

plugins/modules/k8s_cp.py
abikouo: yes indeed, done

abikouo:
recheck

abikouo:
@alinabuzachis @gravesm the build succeeded and #380 has been created to track support for check_mode.
Please add your final review or comment if something is missing

abikouo:
regate

abikouo:
regate

abikouo:
regate

abikouo:
recheck

abikouo:
regate

Issue #244, opened 2021-09-22
k8s_scale does not support check mode


gravesm:
##### SUMMARY

The k8s_scale module claims to support check mode, but it does not. In this case, trying to use check mode results in an exception because the `result` variable in the following block is not bound: https://github.com/ansible-collections/kubernetes.core/blob/a27c701afe597aff6da65331bcdbb70b529a67f3/plugins/modules/k8s_scale.py#L267-L270

In addition to fixing the bug, we should add a test for check mode.

See also #243 
##### ISSUE TYPE
- Bug Report

##### ANSIBLE VERSION

```paste below

```

##### COLLECTION VERSION

```paste below

```

##### CONFIGURATION

```paste below

```

##### STEPS TO REPRODUCE

```yaml

```

##### ACTUAL RESULTS

```paste below

```

PR #255, opened 2021-10-06
k8s_scale, k8s_rollback - add support for check_mode 

### LINKED ISSUE #243: k8s_rollback does not support check_mode
##### SUMMARY

The k8s_rollback module claims to support check mode, but it does not. This should be easy to do with dry run, but since we are still supporting older kubernetes client versions we will also need a non-dry run solution.

We should also add a check_mode test.
##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

k8s_rollback
##### ANSIBLE VERSION

```paste below

```

##### COLLECTION VERSION

```paste below

```

##### CONFIGURATION

```paste below

```

##### STEPS TO REPRODUCE

```yaml

```

##### ACTUAL RESULTS

```paste below

```

abikouo:
##### SUMMARY

closes #243 and #244 

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

k8s_scale
k8s_rollback
##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/molecule/default/tasks/rollback.yml
+++ b/molecule/default/tasks/rollback.yml
@@ -8,18 +8,12 @@
       k8s:
         name: "{{ namespace }}"
         kind: Namespace
-        api_version: v1
-        apply: no
-      register: output
-
-    - name: show output
-      debug:
-        var: output
 
     - name: Create a deployment
       k8s:
         state: present
         wait: yes
+        wait_timeout: 30
         inline: &deploy
           apiVersion: apps/v1
           kind: Deployment
@@ -43,11 +37,7 @@
                     image: nginx:1.17
                     ports:
                       - containerPort: 80
-      register: output
 
-    - name: Show output
-      debug:
-        var: output
 
     - name: Crash the existing deployment
       k8s:
@@ -77,7 +67,47 @@
                     ports:
                       - containerPort: 80
       ignore_errors: yes
-      register: output
+      register: crash
+
+    - name: Assert that the Deployment failed
+      assert:
+        that:
+          - crash is failed
+
+    - name: Read the deployment
+      k8s_info:
+        kind: Deployment
+        name: nginx-deploy
+        namespace: "{{ namespace }}"
+      register: deployment
+
+    - set_fact:
+        failed_version: "{{ deployment.resources[0].metadata.annotations['deployment.kubernetes.io/revision'] }}"
+
+    - name: Rolling Back the crashed deployment (check mode)
+      k8s_rollback:
+        api_version: apps/v1
+        kind: Deployment
+        name: nginx-deploy
+        namespace: "{{ namespace }}"
+      register: result
+      check_mode: yes
+
+    - assert:
+        that:
+          - result is changed
+
+    - name: Read the deployment
+      k8s_info:
+        kind: Deployment
+        name: nginx-deploy
+        namespace: "{{ namespace }}"
+      register: deployment
+
+    - name: Validate that Rollback using check_mode did not changed the Deployment
+      assert:
+        that:
+          - failed_version == deployment.resources[0].metadata.annotations['deployment.kubernetes.io/revision']
 
     - name: Rolling Back the crashed deployment
       k8s_rollback:
@@ -85,12 +115,24 @@
         kind: Deployment
         name: nginx-deploy
         namespace: "{{ namespace }}"
-      when: output.failed
-      register: output
+      register: result
 
-    - name: Show output
-      debug:
-        var: output
+    - name: assert rollback is changed
+      assert:
+        that:
+          - result is changed
+
+    - name: Read the deployment once again
+      k8s_info:
+        kind: Deployment
+        name: nginx-deploy
+        namespace: "{{ namespace }}"
+      register: deployment
+
+    - name: Validate that Rollback changed the Deployment
+      assert:
+        that:
+          - failed_version | int + 1 == deployment.resources[0].metadata.annotations['deployment.kubernetes.io/revision'] | int
 
     - name: Create a DaemonSet
       k8s:
@@ -139,16 +181,12 @@
                   - name: varlibdockercontainers
                     hostPath:
                       path: /var/lib/docker/containers
-      register: output
-
-    - name: Show output
-      debug:
-        var: output
 
     - name: Crash the existing DaemonSet
       k8s:
         state: present
         wait: yes
+        wait_timeout: 30
         definition:
           apiVersion: apps/v1
           kind: DaemonSet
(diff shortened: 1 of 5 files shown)


abikouo:
recheck

Akasurde:
recheck

plugins/modules/k8s_scale.py
```diff
@@ -271,22 +271,25 @@ def _continue_or_exit(warn):
...
+                    module,
+                    k8s_ansible_mixin,
+                    resource,
+                    existing,
+                    replicas,
+                    wait,
+                    wait_time,
+                    wait_sleep,
+                )
+                changed = changed or result["changed"]
```
alinabuzachis, line 292: I guess you can do something like this if you want` changed |= result["changed"]`

molecule/default/tasks/scale.yml
```diff
@@ -45,6 +45,37 @@
         field_selectors:
           - status.phase=Running
 
+    - name: Scale the deployment (check_mode)
```
alinabuzachis, line 48: What about also trying to verify the idempotency overall?

plugins/modules/k8s_scale.py
abikouo: yes we can do that but this is not like a requirement depending on the reviwer people may like this one
So I keep it like that for now

molecule/default/tasks/scale.yml
abikouo: This is was not part of the change, I am just validating the check_mode
Idempotency should be addressed in another PR

alinabuzachis: Generally in the aws collections we used to add tasks for idempotency even for check_mode.

molecule/default/tasks/rollback.yml
```diff
@@ -77,20 +67,72 @@
...
+              - result is changed
+
+        - name: Read the deployment once again
+          k8s_info:
+            kind: Deployment
+            name: nginx-deploy
+            namespace: "{{ namespace }}"
+          register: deployment
+
+        - name: Validate that Rollback using check_mode did not changed the Daemonset
```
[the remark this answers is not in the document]
abikouo: The task is correct, this task is used to validate that the deployment has not changed not the result of k8s_rollback executed using check_mode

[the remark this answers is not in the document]
alinabuzachis: Yes, but shouldn't it be something like this? `Validate that Rollback did not change the Deployment`

molecule/default/tasks/scale.yml
alinabuzachis: @gravesm What do you suggest?

plugins/modules/k8s_scale.py
```diff
@@ -271,22 +271,25 @@ def _continue_or_exit(warn):
             continue
 
         if existing_count != replicas:
-            if not module.check_mode:
-                if module.params["kind"] == "job":
-                    existing.spec.parallelism = replicas
-                    result = resource.patch(existing.to_dict()).to_dict()
+            if module.params["kind"] == "job":
```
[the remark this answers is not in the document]
abikouo: this is user input, as there is no contraints associated to that input, I will lowercase this information to avoid any issue

molecule/default/tasks/scale.yml
abikouo: yes but in AWS collection you do not check the status of the resource, so checking the deployment after running the task using check_mode is another way to test the idempotency

alinabuzachis: I don't think I get your point, but since the task in run in check_mode we don't expect to have any real change on the resource. With the idempotency in general, I mean running the same task twice.

abikouo: by the way it is done

molecule/default/tasks/rollback.yml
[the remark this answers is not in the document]
alinabuzachis: @abikouo Can you please confirm this? Just to be sure.

molecule/default/tasks/rollback.yml
```diff
@@ -77,20 +67,72 @@
...
+              - result is changed
+
+        - name: Read the deployment once again
+          k8s_info:
+            kind: Deployment
+            name: nginx-deploy
+            namespace: "{{ namespace }}"
+          register: deployment
+
+        - name: Validate that Rollback using check_mode did not changed the Deployment
```
alinabuzachis, line 129: But the previous task is no longer performed with check_mode https://github.com/ansible-collections/kubernetes.core/pull/255/files#diff-f71da7f1f86c5838171a1e730fbf7185ceae88d529d4455cf26c74b069291e18R109. I would probably expect to have a change. For this reason I thought the name of the tasks should probably be rephrased. But maybe I am wrong.

molecule/default/tasks/rollback.yml
```diff
@@ -77,20 +67,72 @@
                     ports:
                       - containerPort: 80
       ignore_errors: yes
-      register: output
+      register: crash
```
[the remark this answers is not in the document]
abikouo: ok I will add an assert to validate that the crash task failed and always run the rollback block

molecule/default/tasks/rollback.yml
abikouo: ok you are right

plugins/modules/k8s_scale.py
```diff
@@ -342,25 +345,32 @@ def scale(
...
+            resource.scale.patch(body=scale_obj)
+        except Exception as exc:
+            module.fail_json(msg="Scale request failed: {0}".format(exc))
+
+        k8s_obj = resource.get(name=name, namespace=namespace).to_dict()
+        match, diffs = k8s_ansible_mixin.diff_objects(existing.to_dict(), k8s_obj)
+        result["result"] = k8s_obj
+        result["changed"] = not match
+        if module._diff:
+            result["diff"] = diffs
```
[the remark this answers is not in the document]
abikouo: we can do that but from what I have seen, ansible cant activate both check_mode and _diff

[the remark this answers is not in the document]
gravesm: diff works fine in check mode. Where are you seeing that it doesn't?

abikouo:
recheck

abikouo:
@alinabuzachis @gravesm The build succeeded and all comments were addressed, could you please add your final review? 
Thanks

alinabuzachis (approved):

abikouo:
regate

PR #264, opened 2021-10-14
K8s_taint new module

alinabuzachis:
##### SUMMARY

`k8s_taint` - new module to apply/remove taints to/from nodes.

##### ISSUE TYPE

- New Module Pull Request

##### COMPONENT NAME

k8s_taint

### DIFF
--- a/plugins/modules/k8s_taint.py
+++ b/plugins/modules/k8s_taint.py
@@ -13,7 +13,7 @@ __metaclass__ = type
 DOCUMENTATION = r"""
 module: k8s_taint
 short_description: Taint a node in a Kubernetes/OpenShift cluster
-version_added: "2.1.0"
+version_added: "2.3.0"
 author: Alina Buzachis (@alinabuzachis)
 description:
     - Taint allows a node to refuse Pod to be scheduled unless that Pod has a matching toleration.
@@ -72,7 +72,6 @@ EXAMPLES = r"""
     taints:
         - effect: NoExecute
           key: "key1"
-          value: "value1"
 
 - name: Taint node "foo"
   kubernetes.core.k8s_taint:
@@ -86,12 +85,6 @@ EXAMPLES = r"""
           key: "key1"
           value: "value1"
 
-- name: Remove all taints from "foo" with "key=key1".
-  kubernetes.core.k8s_taint:
-    state: absent
-    name: foo
-    taints: "key1"
-
 - name: Remove taint from "foo".
   kubernetes.core.k8s_taint:
     state: absent
@@ -104,17 +97,37 @@ EXAMPLES = r"""
 
 RETURN = r"""
 result:
-   description:
+    description:
         -  Shows if the node has been successfully tainted/untained.
-   type: str
-   returned: success
-   sample: "node/ip-10-0-152-161.eu-central-1.compute.internal untainted"
+    returned: success
+    type: complex
+    contains:
+        api_version:
+            description: The versioned schema of this representation of an object.
+            returned: success
+            type: str
+        kind:
+            description: Represents the REST resource this object represents.
+            returned: success
+            type: str
+        metadata:
+            description: Standard object metadata. Includes name, namespace, annotations, labels, etc.
+            returned: success
+            type: complex
+        spec:
+            description: Specific attributes of the object. Will vary based on the I(api_version) and I(kind).
+            returned: success
+            type: complex
+        status:
+            description: Current status details for the object.
+            returned: success
+            type: complex
 """
-
 
 import copy
 
 from ansible.module_utils.basic import AnsibleModule
+
 from ansible_collections.kubernetes.core.plugins.module_utils.args_common import (
     AUTH_ARG_SPEC,
 )
@@ -140,55 +153,6 @@ def _get_difference(a, b):
     return [
         a_item for a_item in a if not any(_equal_dicts(a_item, b_item) for b_item in b)
     ]
-
-
-def _get_intersection(a, b):
-    return [a_item for a_item in a if any(_equal_dicts(a_item, b_item) for b_item in b)]
-
-
-def if_failed(function):
-    def wrapper(self, *args):
-        result = {}
-        res = _get_difference(*args)
-        if (
-            not res
-            and self.module.params.get("state") == "present"
-            and not self.module.params.get("overwrite")
-        ):
-            result["result"] = "node/{0} already has ".format(
-                self.module.params.get("name")
-            ) + "{0} taint(s) with same effect(s) and overwrite is false".format(
-                ", ".join(map(lambda t: "%s" % t["key"], _get_intersection(*args),))
-            )
-
-            self.module.exit_json(changed=self.changed, **result)
-
-        if res and self.module.params.get("state") == "absent":
-            self.module.fail_json(
-                msg="{0} not found".format(
-                    ", ".join(map(lambda t: "%s" % t["key"], res))
-                )
-            )
-
-        return function(self, *args)
-
-    return wrapper
-
-
-def if_check_mode(function):
-    def wrapper(self, *args):
-        result = {}
-        if self.module.check_mode:
-            self.changed = True
-            result[
-                "result"
-            ] = "Would have tainted node/{0} if not in check mode".format(
-                self.module.params.get("name")
-            )
-            self.module.exit_json(changed=self.changed, **result)
-        return function(self, *args)
-
-    return wrapper
 
 
 def argspec():
@@ -224,12 +188,11 @@ class K8sTaintAnsible:
         self.k8s_ansible_mixin.exit_json = self.module.exit_json
         self.k8s_ansible_mixin.warn = self.module.warn
         self.k8s_ansible_mixin.warnings = []
-
         self.api_instance = core_v1_api.CoreV1Api(self.k8s_ansible_mixin.client.client)
         self.k8s_ansible_mixin.check_library_version()
         self.changed = False
 
-    def get_current_taints(self, name):
+    def get_node(self, name):
         try:
             node = self.api_instance.read_node(name=name)
         except ApiException as exc:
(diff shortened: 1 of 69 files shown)


alinabuzachis:
recheck

alinabuzachis:
recheck

alinabuzachis:
recheck

alinabuzachis:
recheck

gravesm (changes requested):

plugins/modules/k8s_taint.py
```diff
@@ -0,0 +1,323 @@
...
+        choices: [ present, absent ]
+    name:
+        description:
+            - The name of the node.
+        required: true
+        type: str
+    taints:
+        description:
+            - List containing the taints.
+        type: list
```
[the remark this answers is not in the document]
alinabuzachis: Yeah, you're right, I was sure I had updated the branch. It was an initial try. I completely removed.

plugins/modules/k8s_taint.py
```diff
@@ -0,0 +1,279 @@
...
+    pass
+
+
+def _equal_dicts(a, b):
+    ignore_keys = ("time_added", "value")
+    ka = set(a).difference(ignore_keys)
+    kb = set(b).difference(ignore_keys)
+
+    return all(a[k] == b[k] for k in ka if k in kb)
+
```
abikouo, line 154:
```suggestion
def _equal_dicts(a, b):
    ignore_keys = ("time_added", "value")
    ka = set(a).difference(ignore_keys)
    kb = set(b).difference(ignore_keys)

    return ka == kb and all(a[k] == b[k] for k in ka if k in kb)

```

seems that this method is returning false positive for 
```
a = {'a': 1, 'b': 2}
b = {'a': 1}
```

plugins/modules/k8s_taint.py
```diff
@@ -0,0 +1,279 @@
...
+        return result.to_dict()
+
+    def execute_module(self):
+        result = {"result": {}}
+        state = self.module.params.get("state")
+        taints = self.module.params.get("taints")
+        name = self.module.params.get("name")
+
+        node = self.get_node(name)
+        existing_taints = node.spec.to_dict()["taints"] or []
```
abikouo, line 236:
```suggestion
        existing_taints = node.spec.to_dict().get("taints", [])
```
This is raising exception when the key does not exist

plugins/modules/k8s_taint.py
```diff
@@ -0,0 +1,279 @@
...
+                else:
+                    result["result"] = self.patch_node(taints=taints)
+
+                self.changed = True
+
+        elif state == "absent":
+            if not existing_taints:
+                result["result"] = node.to_dict()
+            if not diff:
+                self.changed = True
```
abikouo, line 262: module execution should not reflect changes when using ``check_mode``

abikouo (dismissed):
Great Job @alinabuzachis!!

plugins/modules/k8s_taint.py
gravesm: I think the function name is probably just a little misleading, but I can't think of a better one. A taint of `{"key": "foo"}` should remove any existing taints with `key=foo` regardless what their value/effect are, so in this case they should be considered "equal enough". I believe this is working correctly, though it would be a good idea to add a few more tests for these kinds of cases.

alinabuzachis: The input won't be in this way.
```
a = {'a': 1, 'b': 2}
b = {'a': 1}
```
The input would always has the same size (for this reason I remove some fields). a and b would always be a combination of {"effect": "NoExecute", "key": "key1"}, so the function works as expected.

plugins/modules/k8s_taint.py
alinabuzachis: In `check_mode` we don't trigger any change, but we only simulate it.

plugins/modules/k8s_taint.py
alinabuzachis: Yeah, like this compact way. Thanks.

plugins/modules/k8s_taint.py
alinabuzachis: Since the key always exists for the moment and if in a near future it won't be returned anymore, I did this for the moment.
`existing_taints = node.spec.to_dict().get("taints") or []`

plugins/modules/k8s_taint.py
```diff
@@ -0,0 +1,275 @@
...
+try:
+    from kubernetes.client.api import core_v1_api
+    from kubernetes.dynamic.exceptions import ApiException
+except ImportError:
+    # ImportError are managed by the common module already.
+    pass
+
+
+def _equal_dicts(a, b):
+    return all([ a[x] == b[x] for x in ('effect', 'key') ])
```
gravesm, line 149: I don't think this change is going to work. You at least need to also look at `value`. In addition, the dictionaries being compared may not have an `effect`. For example, it should be possible to specify the following:
```
state: absent
taints:
  - key: foo
```
This should remove any taint with the key `foo`.

gravesm: So I'm wrong about the value, as that does not seem to matter. We do still need to support deleting by keys. I feel like something like this would work:
```python
def clean(d):
    return {k: v for k, v in d.items() if k in ("key", "effect")}


def within(a, b):
    return clean(a).items() <= clean(b).items()


def difference(a, b):
    diff = []
    for x in b:
        for y in a:
            if not within(x, y):
                diff.append(y)
    return diff
```

alinabuzachis: Yes, we don't need the value for deleting, but only for updating. I'm actually finishing the tests since I added more and I'll push the changed in a while. Thanks.

alinabuzachis:
recheck

alinabuzachis:
recheck

alinabuzachis:
recheck

gravesm:
recheck

gravesm:
recheck

alinabuzachis:
recheck

gravesm:
There are a couple things here that don't seem to be working quite right. The `overwrite` behavior does not match what the documentation describes, since changing the same taint works fine without it:

```
- kubernetes.core.k8s_taint:
    name: "{{ node }}"
    taints:
      - key: key1
        effect: PreferNoSchedule

- kubernetes.core.k8s_taint:
    name: "{{ node }}"
    taints:
      - key: key1
        effect: PreferNoSchedule
        value: val1
```
```
$ kubectl get node kind-control-plane -o jsonpath="{.spec.taints}" | jq .
[
  {
    "effect": "PreferNoSchedule",
    "key": "key1",
    "value": "val1"
  }
]
```
The more I think about it, the more I think we should not be trying to replicate the exact behavior of `kubectl taint`. I actually think the way the module works now is more appropriate to the desired state model of Ansible. My only suggestions would be to rename `overwrite` to `replace` to avoid any confusion with `kubectl`, and to update the documentation to match the behavior. In other words, the default behavior should update any existing taints and append any new taints. With `replace` it should replace all the existing taints with the provided taints.

The delete behavior also seems to not always work:
```
- kubernetes.core.k8s_taint:
    name: "{{ node }}"
    taints:
      - key: key1
        effect: PreferNoSchedule

- kubernetes.core.k8s_taint:
    name: "{{ node }}"
    state: absent
    taints:
      - key: key1
      - key: key2
```
```
$ kubectl get node kind-control-plane -o jsonpath="{.spec.taints}" | jq .
[
  {
    "effect": "PreferNoSchedule",
    "key": "key1"
  }
]
```

alinabuzachis:
> There are a couple things here that don't seem to be working quite right. The `overwrite` behavior does not match what the documentation describes, since changing the same taint works fine without it:
> 
> ```
> - kubernetes.core.k8s_taint:
>     name: "{{ node }}"
>     taints:
>       - key: key1
>         effect: PreferNoSchedule
> 
> - kubernetes.core.k8s_taint:
>     name: "{{ node }}"
>     taints:
>       - key: key1
>         effect: PreferNoSchedule
>         value: val1
> ```
> 
>  ```
> $ kubectl get node kind-control-plane -o jsonpath="{.spec.taints}" | jq .
> [
>   {
>     "effect": "PreferNoSchedule",
>     "key": "key1",
>     "value": "val1"
>   }
> ]
> ```
> 
> The more I think about it, the more I think we should not be trying to replicate the exact behavior of `kubectl taint`. I actually think the way the module works now is more appropriate to the desired state model of Ansible. My only suggestions would be to rename `overwrite` to `replace` to avoid any confusion with `kubectl`, and to update the documentation to match the behavior. In other words, the default behavior should update any existing taints and append any new taints. With `replace` it should replace all the existing taints with the provided taints.
> 
> The delete behavior also seems to not always work:
> 
> ```
> - kubernetes.core.k8s_taint:
>     name: "{{ node }}"
>     taints:
>       - key: key1
>         effect: PreferNoSchedule
> 
> - kubernetes.core.k8s_taint:
>     name: "{{ node }}"
>     state: absent
>     taints:
>       - key: key1
>       - key: key2
> ```
> 
>  ```
> $ kubectl get node kind-control-plane -o jsonpath="{.spec.taints}" | jq .
> [
>   {
>     "effect": "PreferNoSchedule",
>     "key": "key1"
>   }
> ]
> ```

The module works in this way at the moment. 

1. If there is already a taint with `key: key1 effect: PreferNoSchedule` and I am applying a new one  `key: key1 effect: PreferNoSchedule value: val1`, then the actual taint is updated with the new value. There is only an update.

2. Concerning overwrite, if there is already a taint with `key: key1 effect: PreferNoSchedule` and I want to apply the same taint with overwrite it won't show any change. If I would like to apply instead `key: key2 effect: PreferNoSchedule` and `overwrite: true`, then the current taints will be only `key: key2 effect: PreferNoSchedule`.

While in the 1. case, I can update different taints at the same time (of course, if there are update to be applied), if does not any taints.

With overwrite, you actually get rid of the current taints and updates them with the ones you want to apply. 


Concerning the delete behaviour, if my current taints are `key: key1` and` key: key2` and if I want to delete `key: key1`, ` key: key2` and `key: key3`, it won't delete any of the taints. There should be an 100% match to have `key: key1` and` key: key2`deleted even if `key: key3` does not exist. This is the behaviour of `kubectl taint node .......`

gravesm:
> The module works in this way at the moment.
> 
>     1. If there is already a taint with `key: key1 effect: PreferNoSchedule` and I am applying a new one  `key: key1 effect: PreferNoSchedule value: val1`, then the actual taint is updated with the new value. There is only an update.
> 
>     2. Concerning overwrite, if there is already a taint with `key: key1 effect: PreferNoSchedule` and I want to apply the same taint with overwrite it won't show any change. If I would like to apply instead `key: key2 effect: PreferNoSchedule` and `overwrite: true`, then the current taints will be only `key: key2 effect: PreferNoSchedule`.
> 
> 
> While in the 1. case, I can update different taints at the same time (of course, if there are update to be applied), in the 2. case, you actually get rid of the current taints and updates them with the ones you want to apply.

Right, what I'm saying is I agree this is the correct behavior, but we should rename `overwrite` to `replace` because the `overwrite` option with `kubectl taint` works differently. People may find this confusing when coming from `kubectl`. We also need to update the documentation because the behavior you describe is not what the documentation describes.
 
> Concerning the delete behaviour, if my current taints are `key: key1` and` key: key2` and if I want to delete `key: key1`, ` key: key2` and `key: key3`, it won't delete any of the taints. There should be an 100% match to have `key: key1` and` key: key2`deleted even if `key: key3` does not exist. This is the behaviour of `kubectl taint node .......`

I think we should diverge from `kubectl taint` here because this behavior requires the user to know too much about the existing state of the cluster. We want someone to be able to express a desired state like "I want the node to not have any taints with key1 or key2." It shouldn't matter what the current state of the node is. By the time the task is finished, any taints with key1 or key2 should have been removed.

alinabuzachis:
recheck

alinabuzachis:
recheck

gravesm:
The idempotency when adding taints doesn't seem to work quite right. If a node already has taints T1 and T2 then the following all works correctly:
```
- taint:
    name: "{{ node }}"
    taints:
      - "{{ T2 }}"
- taint:
    name: "{{ node }}"
    taints:
      - "{{ T1 }}"
      - "{{ T2 }}"
      - "{{ T3 }}"
```
but then the following fails:
```
- taint:
    name: "{{ node }}"
    taints:
      - "{{ T1 }}"
      - "{{ T4 }}"
```
I would expect the module by default to be additive. So if the node already has taints T1, T2 and T3, if I tell it to add T1 and T4 then the end result should be T1, T2, T3, T4.

The delete and replace functionality works as expected.

alinabuzachis:
> The idempotency when adding taints doesn't seem to work quite right. If a node already has taints T1 and T2 then the following all works correctly:
> 
> ```
> - taint:
>     name: "{{ node }}"
>     taints:
>       - "{{ T2 }}"
> - taint:
>     name: "{{ node }}"
>     taints:
>       - "{{ T1 }}"
>       - "{{ T2 }}"
>       - "{{ T3 }}"
> ```
> 
> but then the following fails:
> 
> ```
> - taint:
>     name: "{{ node }}"
>     taints:
>       - "{{ T1 }}"
>       - "{{ T4 }}"
> ```
> 
> I would expect the module by default to be additive. So if the node already has taints T1, T2 and T3, if I tell it to add T1 and T4 then the end result should be T1, T2, T3, T4.
> 
> The delete and replace functionality works as expected.

The actual behaviour of the module didn't allow to apply an already present taint, for this reason it was failing. The module itself has an additive behaviour, but didn't allow you to add the same taint twice. It should be fixed now.

gravesm:
recheck

gravesm:
recheck

PR #272, opened 2021-10-26
add no_proxy support to k8s*

### LINKED ISSUE #271: add no_proxy support to k8s*
##### SUMMARY

current kubernetes.core supports parameter for proxy but NOT no_proxy.
as the result, operator needs to set/unset proxy according to the target clusters.
it is bothering task because kubectl supports both proxy and no_proxy environment variables.

no_proxy is supported in latest kubernetes python library (since kubernetes>=19.15.0).
so lets support it in kubernetes.core for ansible.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

plugins/module_utils/args_common.py
plugins/modules/k8s*

##### ADDITIONAL INFORMATION

```yaml
  - k8s:
      state: present
      src: "deployment.yaml"
      proxy:      "http://proxy.yourdomain.com:8080/"
      no_proxy:   "localhost,.yourdomain.com,127.0.0.0/8,10.0.0.0/8,172.16.0.0/12,192,168.0.0/16"
```
or, set environment variables K8S_AUTH_NO_PROXY as well as K8S_AUTH_PROXY

itaru2622:
##### SUMMARY

close #271

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

plugins/module_utils/args_common.py
plugins/modules/k8s*

##### ADDITIONAL INFORMATION

It requires latest kubernetes library(>=19.15.0) to use this feature.

```paste below

pip install kubernetes>=19.15.0
then, use following snippet yaml:

  - k8s:
      state: present
      src: "deployment.yaml"
      proxy:      "http://proxy.yourdomain.com:8080/"
      no_proxy:   "localhost,.yourdomain.com,127.0.0.0/8,10.0.0.0/8,172.16.0.0/12,192,168.0.0/16"

or use environment variable K8S_AUTH_NO_PROXY as well as K8S_AUTH_PROXY.
```

### DIFF
--- a/plugins/module_utils/common.py
+++ b/plugins/module_utils/common.py
@@ -212,6 +212,13 @@ def get_api_client(module=None, **kwargs):
         # Removing trailing slashes if any from hostname
         auth["host"] = auth.get("host").rstrip("/")
 
+    if auth_set("no_proxy"):
+        if LooseVersion(kubernetes.__version__) < LooseVersion("19.15.0"):
+            _raise_or_fail(
+                Exception("kubernetes >= 19.15.0 is required to use no_proxy feature."),
+                "Failed to set no_proxy due to: %s",
+            )
+
     if auth_set("username", "password", "host") or auth_set("api_key", "host"):
         # We have enough in the parameters to authenticate, no need to load incluster or kubeconfig
         pass
--- a/changelogs/fragments/272-k8s-add-support-no_proxy.yaml
+++ b/changelogs/fragments/272-k8s-add-support-no_proxy.yaml
@@ -1,0 +1,2 @@
+minor_changes:
+  - k8s - add no_proxy support to k8s* (https://github.com/ansible-collections/kubernetes.core/pull/272).
--- a/plugins/doc_fragments/k8s_auth_options.py
+++ b/plugins/doc_fragments/k8s_auth_options.py
@@ -84,7 +84,7 @@ options:
     - This feature requires kubernetes>=19.15.0. When kubernetes library is less than 19.15.0, it fails even no_proxy set in correct.
     - example value is "localhost,.local,.example.com,127.0.0.1,127.0.0.0/8,10.0.0.0/8,172.16.0.0/12,192.168.0.0/16"
     type: str
-    version_added: 2.2.2
+    version_added: 2.3.0
   proxy_headers:
     description:
     - The Header used for the HTTP proxy.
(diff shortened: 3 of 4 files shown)


abikouo (dismissed):
@itaru2622 I just forgot to mention that you will also need a change log to report this change.

itaru2622:
@abikouo your comment applied. please review it.  at current, some CI tests were failed but I could imagine what should I do.

was it caused by 'version_added:2.2.2'  or do I need to fix?
I cannot imagine what is wrong because of less message in failure reason. Is there any way to test them in local?

gravesm:
@itaru2622 Since this is a new feature and not a bugfix, the version_added needs to be 2.3.0.

itaru2622:
@gravsm Thank you for your comment. I updated version_added to 2.3.0.

gravesm:
recheck

itaru2622:
@gravesm thank you for your help.

itaru2622:
@gravesm @abikouo finally first CI test passed, what should I do next? could you review this, please.

gravesm:
@itaru2622 The second CI check will run once the PR is approved. Could you add a failure when the kubernetes version is not high enough? You can add it after the `for` block here if the `no_proxy` param has been set: https://github.com/ansible-collections/kubernetes.core/blob/91b80b1d1d99297090649778e89e172a13746252/plugins/module_utils/common.py#L181 See https://github.com/ansible-collections/kubernetes.core/blob/91b80b1d1d99297090649778e89e172a13746252/plugins/module_utils/common.py#L198 for how to do the failure in this function.

itaru2622:
@gravesm thank you I will apply them

itaru2622:
@gravesm I applied your comment, please check it.

plugins/module_utils/common.py
```diff
@@ -178,6 +178,13 @@ def _raise_or_fail(exc, msg):
             if env_value is not None:
                 if AUTH_ARG_SPEC[arg_name].get("type") == "bool":
                     env_value = env_value.lower() not in ["0", "false", "no"]
+                if true_name == 'no_proxy' and LooseVersion(kubernetes.__version__) < LooseVersion("19.15.0"):
```
gravesm, line 181: Could you move this down to line 221, after the `if auth_set("host")` block? It's currently only checking if NO_PROXY has been set through the environment variable. It needs to check against the `auth` dict after we're done creating it. So something like:
```
if auth_set("no_proxy") and LooseVersion...
```

gravesm (changes requested):

itaru2622:
@gravesm I applied your comment about kubernetes version check, and rebased to latest main branch.

gravesm (approved):
Thanks for your work on this!

gravesm:
recheck

gravesm:
recheck

gravesm:
recheck

gravesm:
recheck

PR #273, opened 2021-10-26
kustomize lookup plugin: build_flags

eb4x:
Allow setting a string of optional build_flags at the end of the
command.

##### SUMMARY

A string to add some params/options to the end of "kustomize build dir" / "kubectl kustomize dir" command.
This could have been a list, or some other grandiose parsing event, I haven't put much thought into other use-cases.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

kustomize lookup plugin

##### ADDITIONAL INFORMATION

It feels like a one-off kind of thing, where I just need to add "--enable-alpha-plugins" at the end. Maybe someone else will end up in a similar predicament later down the line, and be relieved that they can just build_flags="--all-my great --build-options are-available --yay".

### DIFF
--- a/plugins/lookup/kustomize.py
+++ b/plugins/lookup/kustomize.py
@@ -30,6 +30,9 @@ DOCUMENTATION = """
       opt_dirs:
         description:
         - An optional list of directories to search for the executable in addition to PATH.
+      build_flags:
+        description:
+        - Set additional build flags, like --enable-alpha-plugins
 
     requirements:
       - "python >= 3.6"
@@ -135,6 +138,9 @@ class LookupModule(LookupBase):
                     executable_path
                 )
             )
+
+        if "build_flags" in kwargs:
+            command += [kwargs["build_flags"]]
 
         (out, err) = run_command(command)
         if err:


abikouo (changes requested):
changelog required
update molecule tests https://github.com/ansible-collections/kubernetes.core/blob/main/molecule/default/tasks/lookup_kustomize.yml

plugins/lookup/kustomize.py
```diff
@@ -30,6 +30,9 @@
       opt_dirs:
         description:
         - An optional list of directories to search for the executable in addition to PATH.
+      build_flags:
+        description:
+        - Set additional build flags, like --enable-alpha-plugins
```
Akasurde, line 35:
```suggestion
        - Set additional build flags, like C(--enable-alpha-plugins).
```

plugins/lookup/kustomize.py
```diff
@@ -136,6 +139,9 @@ def run(
                 )
             )
 
+        if "build_flags" in kwargs:
```
Akasurde, line 142: Please create a [changelog fragment](https://docs.ansible.com/ansible/latest/community/development_process.html#creating-a-changelog-fragment). See [this fragment](https://github.com/ansible/ansible/blob/c2f528b7686c9573aa8493357219e0bbe4416dd3/changelogs/fragments/reboot-change-default-boot-command.yaml) as an example.

Akasurde (changes requested):

PR #288, opened 2021-11-18
add an example to the kubectl plugin

tpo:
##### SUMMARY

When looking at the [kubectl plugin](https://docs.ansible.com/ansible/latest/collections/community/kubernetes/kubectl_connection.html) documentation I have *no* idea how to use the plugin. Having and example helps *a lot*. Please include an example such as the provided one. Thanks a lot for writing the plugin in the first place!

##### ISSUE TYPE

- Docs Pull Request

##### COMPONENT NAME

kubectl

### DIFF
--- a/plugins/connection/kubectl.py
+++ b/plugins/connection/kubectl.py
@@ -171,6 +171,37 @@ DOCUMENTATION = r"""
         aliases: [ kubectl_verify_ssl ]
 """
 
+# example taken and adapted from
+# https://github.com/geerlingguy/ansible-for-devops/blob/master/kubernetes/examples/kubectl-connection.yml
+#
+EXAMPLES = r"""
+# This playbook assumes you already have the kubectl binary installed
+# and available in the $PATH.
+- hosts: k8s-master
+  become: yes
+
+  tasks:
+    - name: Get the phpmyadmin Pod name.
+      command: >
+        kubectl --no-headers=true get pod -l app=phpmyadmin
+        -o custom-columns=:metadata.name
+      register: phpmyadmin_pod
+
+    - name: Add the phpmyadmin Pod to the inventory.
+      add_host:
+        name: '{{ phpmyadmin_pod.stdout }}'
+        ansible_kubectl_namespace: default
+        ansible_connection: kubectl
+
+    # Note: Python is required to use other modules.
+    - name: Run a command inside the container.
+      shell: date
+      register: date_output
+      delegate_to: '{{ phpmyadmin_pod.stdout }}'
+
+    - debug: var=date_output.stdout
+"""
+
 import distutils.spawn
 import os
 import os.path


gravesm:
recheck

plugins/connection/kubectl.py
```diff
@@ -171,6 +171,37 @@
...
+#
+EXAMPLES = r"""
+# This playbook assumes you already have the kubectl binary installed
+# and available in the $PATH.
+- hosts: k8s-master
+  become: yes
+
+  tasks:
+    - name: Get the phpmyadmin Pod name.
+      command: >
```
gravesm, line 185: Thanks for the PR @tpo. Instead of using `kubectl` here, can you instead use the `kubernetes.core.k8s_info` module?

gravesm:
recheck

plugins/connection/kubectl.py
tpo: @gravesm wrote:

> Instead of using kubectl here, can you instead use the kubernetes.core.k8s_info module?

I'd like to. I just spent 90 minutes trying to get a simple `k8s_info` task to run to no avail. Same as with the kubectl plugin: the examples in the `k8s_info` docu are too dire to allow me to get it to work.

* Either `k8s_info` tries to first connect to the master node via ssh and access the k8s API there, which fails, because the credentials to access the API are not available on the node because my management machine is my own PC and not the master node.
* Or I am trying to get accessing the API from my local PC to work:

```
- hosts: master-node
  tasks:
    - local_action:
        module: k8s_info
        kind: Pod
        context: my_cluster
        # the following line doesn't seem to have any impact -
        # ansible will still try to connect to the k8s API on localhost
        # kubeconfig: /home/myself/.kube/config
      become: no
      register: pod_list
    - debug: var=pod_list
```
but am failing because for whatever reason, `k8s_info` is **not** accessing my default ~/.kube/config but instead trying to connect to the k8s API at localhost, which obviously fails, because localhost is not in fact a k8s cluster. Nota bene I'm running ansible 2.10 on Debian bullseye here.

So for the time being I'm stopping trying to get k8s_info to work and instead returning to my usual paid job duties again.

That means there are the following ways forward for me with this ticket:

* I get help on how to get `k8s_info` to work
* at some undefined time in the future I again take some time off from my normal schedule and try to resolve this twisted `k8s_info` mystery
* we leave the example as is
* I drop the retrieval of pod names from the example (which will obviously make it much less useful - because pod names are usually ephemeral) and put a static pod name into the example instead
* we scratch this pull request
* some other idea

What do you think? Do you have help or a suggestion on how to proceed?

tpo: Unfortunately it seems that I am unable to reserve myself time for b) above. Any opinions on unstalling this pull request with either choice of a) c) d) e) or f) from above?

Akasurde:
recheck

Akasurde:
recheck

yurnov:
I have experience with `kubernetes.core.kubectl` connection plugin in production, I can create a better example if it help. I agree that looking into current documentation (i.e. for 3.0.0 or 3.1.0 tags) it's not clear how to use connection plugin at all.

tpo:
> I have experience with `kubernetes.core.kubectl` connection plugin in production, I can create a better example if it help.

Thank you, please do

yurnov:
Hi @tpo and @gravesm, would you please review #741 that is my attempt to improve collection documentation by adding examples of using the collection plugin. Comment of @gravesm to use `k8s_info` considered

tpo:
> Hi @tpo and @gravesm, would you please review #741 that is my attempt to improve collection documentation by adding examples of using the collection plugin. Comment of @gravesm to use `k8s_info` considered

Thanks a lot for your very nice examples @yurnov - I have [reviewed](https://github.com/ansible-collections/kubernetes.core/pull/741#issuecomment-2143795916) your pull request. If there's some formal syntax to tag your pull request as reviewed by myself then let me know and I can add that as well.

Thanks again, very appreciated!

yurnov:
Hi @tpo,

I think that you can close this PR as #741 is merged

tpo:
> I think that you can close this PR as #741 is merged

Thanks a **lot** @yurnov - closing this pull request in favor of #741

PR #290, opened 2021-11-19
Return diff in helm check mode

jfrabaute:
When the helm module is executed in check mode with the helm diff plugin
installed, it now returns the diff.

##### SUMMARY
When the helm module is executed in check mode with the helm diff plugin
installed, it now returns the diff.

##### COMPONENT NAME
helm

### DIFF
--- a/plugins/modules/helm.py
+++ b/plugins/modules/helm.py
@@ -537,7 +537,7 @@ def helmdiff_check(
             cmd += " -f=" + values_file
 
     rc, out, err = run_helm(module, cmd)
-    return len(out.strip()) > 0
+    return (len(out.strip()) > 0, out.strip())
 
 
 def default_check(release_status, chart_info, values=None, values_files=None):
@@ -725,7 +725,7 @@ def main():
         else:
 
             if has_plugin(helm_cmd_common, "diff") and not chart_repo_url:
-                would_change = helmdiff_check(
+                (would_change, prepared) = helmdiff_check(
                     module,
                     helm_cmd_common,
                     release_name,
@@ -735,6 +735,8 @@ def main():
                     chart_version,
                     replace,
                 )
+                if would_change and module._diff:
+                    opt_result["diff"] = {"prepared": prepared}
             else:
                 module.warn(
                     "The default idempotency check can fail to report changes in certain cases. "
--- a/changelogs/fragments/290-returns-diff-in-check-mode.yaml
+++ b/changelogs/fragments/290-returns-diff-in-check-mode.yaml
@@ -1,0 +1,2 @@
+minor_changes:
+  - helm - when ansible is executed in check mode, return the diff between what's deployed and what will be deployed.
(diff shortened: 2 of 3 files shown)


plugins/modules/helm.py
```diff
@@ -537,7 +537,7 @@ def helmdiff_check(
             cmd += " -f=" + values_file
 
     rc, out, err = run_helm(module, cmd)
-    return len(out.strip()) > 0
+    return (len(out.strip()) > 0, out.strip())
```
Akasurde, line 540: Please create a [changelog fragment](https://docs.ansible.com/ansible/latest/community/development_process.html#creating-a-changelog-fragment). See [this fragment](https://github.com/ansible/ansible/blob/c2f528b7686c9573aa8493357219e0bbe4416dd3/changelogs/fragments/reboot-change-default-boot-command.yaml) as an example. 

Also, could you please add a test for this change?

jfrabaute: > Please create a [changelog fragment](https://docs.ansible.com/ansible/latest/community/development_process.html#creating-a-changelog-fragment). See [this fragment](https://github.com/ansible/ansible/blob/c2f528b7686c9573aa8493357219e0bbe4416dd3/changelogs/fragments/reboot-change-default-boot-command.yaml) as an example.

Done.

> Also, could you please add a test for this change?

I'll work on this. Is there an example of a helm module test I can use as a base/reference/template?

Thanks.

plugins/modules/helm.py
```diff
@@ -735,6 +735,8 @@ def main():
                     chart_version,
                     replace,
                 )
+                if would_change:
```
gravesm, line 738:
```suggestion
                if would_change and module._diff:
```

plugins/modules/helm.py
gravesm: @jfrabaute You can add some tests in https://github.com/ansible-collections/kubernetes.core/blob/main/molecule/default/roles/helm/tasks/tests_helm_diff.yml. What we'd like to test is that when `diff: yes` is specified in the task that the diff is in the output and when `diff: no` is specified, it's not. You may be able to just modify some of the existing tests, but if not that file should work as a guide for adding some new ones.

plugins/modules/helm.py
jfrabaute: Ansible takes care of displaying the diff or not, so the fact to pass the diff or not here is not changing the output.
So in order to keep it simple, I didn't add this condition. Is that ok?

plugins/modules/helm.py
jfrabaute: Thanks! I'll work on this.

plugins/modules/helm.py
gravesm: Unfortunately, without this the diff will still display if the task has `diff: no` and `--diff` was passed on the command line.

plugins/modules/helm.py
jfrabaute: Tests added

Akasurde:
recheck

gravesm:
recheck

gravesm (approved):

gravesm:
Thanks @jfrabaute!

Issue #126, opened 2021-06-09
Module defaults for k8s group using community.kubernetes collection


erikgb:
##### SUMMARY
I am trying to migrate our Ansible stuff to use FQN on module names (including migrating to `community.kubernetes` collection). Since we are using quite a few of the k8s modules in our playbooks, we have set module defaults for the k8s group, as explained [here](https://docs.ansible.com/ansible/latest/user_guide/playbooks_module_defaults.html#module-defaults-groups). I have made a few attempts to continue using module group defaults, but no luck so far.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
Seems to be a general issue with most modules in the `community.kubernetes` collection.

##### ANSIBLE VERSION

```paste below
ansible@abda10c4cd81:~$ ansible --version
ansible [core 2.11.1]
  config file = None
  configured module search path = ['/home/ansible/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/local/lib/python3.9/site-packages/ansible
  ansible collection location = /home/ansible/.ansible/collections:/usr/share/ansible/collections
  executable location = /usr/local/bin/ansible
  python version = 3.9.5 (default, May 12 2021, 15:26:36) [GCC 8.3.0]
  jinja version = 3.0.1
  libyaml = True
ansible@abda10c4cd81:~$
```

##### CONFIGURATION

```paste below
ansible@abda10c4cd81:~$ ansible-config dump --only-changed
ANSIBLE_FORCE_COLOR(env: ANSIBLE_FORCE_COLOR) = True
DEFAULT_LOAD_CALLBACK_PLUGINS(env: ANSIBLE_LOAD_CALLBACK_PLUGINS) = True
DEFAULT_STDOUT_CALLBACK(env: ANSIBLE_STDOUT_CALLBACK) = yaml
HOST_KEY_CHECKING(env: ANSIBLE_HOST_KEY_CHECKING) = False
INTERPRETER_PYTHON(env: ANSIBLE_PYTHON_INTERPRETER) = auto
ansible@abda10c4cd81:~$
```

##### OS / ENVIRONMENT

Container image based on official Python 3 image, running CI/CD with GitLab CI/CD.
````
ansible@abda10c4cd81:~$ cat /etc/os-release
PRETTY_NAME="Debian GNU/Linux 10 (buster)"
NAME="Debian GNU/Linux"
VERSION_ID="10"
VERSION="10 (buster)"
[...]

PR #296, opened 2021-11-30
Add integration test to check handling of module_defaults 

mandar242:
##### SUMMARY

Add integration test to make sure that ```module_defaults``` are handled correctly in tasks.

Related to #126.

##### ISSUE TYPE

- Bugfix Pull Request

### DIFF
--- a/molecule/default/tasks/full.yml
+++ b/molecule/default/tasks/full.yml
@@ -429,25 +429,23 @@
     # test setting module defaults for community.kubernetes.k8s_info
     - block:
         - name: Create a namespace
-          k8s:
+          kubernetes.core.k8s:
             name: test-namespace-module-defaults
             kind: Namespace
           register: output
 
         - name: Create a ConfigMap
-          k8s:
+          kubernetes.core.k8s:
             kind: ConfigMap
             name: test-configmap-1
-            namespace: test-namespace-module-defaults
             definition:
               data:
                 key1: value1
 
         - name: Create another ConfigMap
-          k8s:
+          kubernetes.core.k8s:
             kind: ConfigMap
             name: test-configmap-2
-            namespace: test-namespace-module-defaults
             definition:
               data:
                 key2: value2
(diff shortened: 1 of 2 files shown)


molecule/default/tasks/full.yml
```diff
@@ -426,6 +426,44 @@
...
+            kind: ConfigMap
+          register: configmap_info
+
+        - assert:
+            that:
+              - configmap_info.resources[1].metadata.name == "test-configmap-1"
+              - configmap_info.resources[2].metadata.name == "test-configmap-2"
+
+      module_defaults:
+        group/k8s:
```
gravesm, line 464: I believe this may need to be `group/kubernetes.core.k8s`

molecule/default/tasks/full.yml
```diff
@@ -426,6 +426,44 @@
...
+        - name: Create another ConfigMap
+          kubernetes.core.k8s:
+            kind: ConfigMap
+            name: test-configmap-2
+            definition:
+              data:
+                key2: value2
+
+        - name: Get list of all ConfigMaps in namespace specified in module_defaults
+          community.kubernetes.k8s_info:
```
gravesm, line 454:
```suggestion
          kubernetes.core.k8s_info:
```

molecule/default/tasks/full.yml
```diff
@@ -426,6 +426,44 @@
...
+            kind: ConfigMap
+          register: configmap_info
+
+        - assert:
+            that:
+              - configmap_info.resources[1].metadata.name == "test-configmap-1"
+              - configmap_info.resources[2].metadata.name == "test-configmap-2"
```
gravesm, line 461: We should check the `metadata.namespace` here, too.

molecule/default/tasks/full.yml
```diff
@@ -426,6 +426,44 @@
         that:
           - result.resources[0].data.testkey == "{{ cmap_data.stdout | b64encode }}"
 
+    # test setting module defaults for community.kubernetes.k8s_info
```
gravesm, line 429:
```suggestion
    # test setting module defaults for kubernetes.core.k8s_info
```

gravesm (commented):
Thanks! A few small things here. Additionally, this block will probably need to only be run when the ansible version is >=2.12. It will likely fail otherwise.

gravesm (approved):

Issue #293, opened 2021-11-26
[Twistlock scan]ansible_collections/kubernetes/core/molecule/default/roles/k8scopy/files/hello security issue


quietsnail:
##### SUMMARY
High security issue in twistlock scan for `/opt/ansible/.ansible/collections/ansible_collections/kubernetes/core/molecule/default/roles/k8scopy/files/hello`

##### ISSUE TYPE
- Bug Report -Security

##### ANSIBLE VERSION

```
RUN ansible --version
ansible [core 2.11.6] 
 config file = /etc/ansible/ansible.cfg
 configured module search path = ['/usr/share/ansible/openshift']
ansible python module location = /usr/local/lib/python3.8/site-packages/ansible
 ansible collection location = /opt/ansible/.ansible/collections:/usr/share/ansible/collections
executable location = /usr/local/bin/ansible
python version = 3.8.8 (default, Aug 11 2021, 06:52:42) [GCC 8.5.0 20210514 (Red Hat 8.5.0-3)]
jinja version = 3.0.1
libyaml = True
```

##### COLLECTION VERSION

```


```

##### CONFIGURATION

```paste below

```

##### STEPS TO REPRODUCE

```yaml

```

##### EXPECTED RESULTS

No twistlock security issues in ansible directory.

##### ACTUAL RESULTS

```
[...]

PR #298, opened 2021-12-01
Remove binary file from molecule test suite

### LINKED ISSUE #297: kubernetes.core ships a binary without coresponding source
##### SUMMARY
`molecule/default/roles/k8scopy/files/hello` is an unstripped go binary for x64_64, which does not come with source code, and does not have the corresponding build rules in place to build it. This unfortunately violates the GPLv3, and makes it undistributable by downstream distributions (in this case Debian).

It would be great to have it also be built by the build command in the Makefile, or have it replaced by something that is distributable. Removing it from the pypi package is also possible as a stop gap solution, but ideally it should ship with all code needed to use the collection, and run the tests, too.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

`molecule/default/roles/k8scopy/files/hello`

##### STEPS TO REPRODUCE

```yaml
file molecule/default/roles/k8scopy/files/hello
molecule/default/roles/k8scopy/files/hello: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), statically linked, Go BuildID=hXQs3Sfk7BP_t5twMGGt/ApiAX9Vfh3MskyxKkiir/fnXuOf8gzVDDuIxFSv-d/HKCGoZO7tghIncxbx8SX, not stripped

```

##### EXPECTED RESULTS

Being able to build said binary from source, e.g. `make build`.

##### ACTUAL RESULTS

Only got the binary without source.

gravesm:
##### SUMMARY

The binary file used to test k8s_cp is causing larger problems
downstream. There's no reason why the binary file needs to function as
all we care about is that the content of the file has not changed during
the copy process. This can be accomplished by comparing file hashes.

Fixes #297 #293

##### ISSUE TYPE

- Bugfix Pull Request

##### ADDITIONAL INFORMATION

```paste below

```


Akasurde (approved):

gravesm:
recheck

nkadel:
I hope it was added to the .gitignore, to avoid accidentally adding it again?

Akasurde:
@nkadel Thanks for the suggestion. This file is part of the test environment and has nothing to do with working on Kubernetes Ansible Modules. @gravesm replaced the binary file with the dynamically generated file and also backported changes to older releases. I hope this does not require adding the binary file name to `.gitignore` file.

PR #308, opened 2021-12-12
Fix for common non-ASCII characters in CRDs

kubealex:
This should keep the module safe from digesting non-ASCII chars like here (https://github.com/projectcalico/api/pull/46/files)

##### SUMMARY
Add support for non-ASCII chars in manifests.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
core.k8s module failing if resources contain non ascii chars


plugins/module_utils/common.py
```diff
@@ -496,7 +496,7 @@ def load_resource_definitions(self, src):
         if not os.path.exists(path):
             self.fail(msg="Error accessing {0}. Does the file exist?".format(path))
         try:
-            with open(path, "r") as f:
+            with open(path, "r", encoding = "utf-8") as f:
```
Akasurde, line 499:
```suggestion
            with open(path, "r", encoding="utf-8") as f:
```

Akasurde: Please create a [changelog fragment](https://docs.ansible.com/ansible/latest/community/development_process.html#creating-a-changelog-fragment). See [this fragment](https://github.com/ansible/ansible/blob/c2f528b7686c9573aa8493357219e0bbe4416dd3/changelogs/fragments/reboot-change-default-boot-command.yaml) as an example.

Akasurde (dismissed):

gravesm (commented):
Thanks for the PR @kubealex. I would suggest instead that we change the `open()` call to:
```python
with open(path, "rb") as f:
```
We can't assume the input file is utf-8 since utf-16 is also a valid encoding for yaml. `yaml.safe_load_all()` can work with a byte stream as well, and already attempts to determine the proper encoding. Any encoding errors will be passed through in the `yaml.YAMLError`.

plugins/module_utils/common.py
kubealex: I created the changelog and also applied @gravesm suggestion to use rb mode for opening the file rather than guessing the UTF-X

gravesm:
I believe the changelog fragment is going to need to have a `.yaml` extension to pass sanity tests.

kubealex:
> I believe the changelog fragment is going to need to have a `.yaml` extension to pass sanity tests.

😄 you're right!

gravesm (approved):
Looks good, thanks!

gravesm:
recheck

gravesm:
recheck

PR #311, opened 2021-12-17
Migrate k8s

alinabuzachis:
Depends-On: https://github.com/ansible/ansible-zuul-jobs/pull/1340

##### SUMMARY

Use refactored `module_utils`.

Depends-On: https://github.com/ansible-collections/cloud.common/pull/102
Depends-On: https://github.com/ansible-collections/kubernetes.core/pull/356
##### COMPONENT NAME

k8s

### DIFF
--- a/plugins/module_utils/k8s/runner.py
+++ b/plugins/module_utils/k8s/runner.py
@@ -44,6 +44,7 @@ def validate(client, module, resource):
 def run_module(module) -> None:
     results = []
     changed = False
+    result = {"changed": changed, "result": {}}
 
     client = get_api_client(module)
     svc = K8sService(client, module)
@@ -61,14 +62,14 @@ def run_module(module) -> None:
             if module.warnings:
                 e["msg"] += "\n" + "\n    ".join(module.warnings)
             if module.params.get("continue_on_error"):
-                result = {"error": "{0}".format(e)}
+                result["error"] = dict(msg="{0}".format(e))
             else:
                 module.fail_json(msg=to_native(e))
 
         if module.warnings:
             result["warnings"] = module.warnings
 
-        changed = changed or result["changed"]
+        changed |= result["changed"]
         results.append(result)
 
     if len(results) == 1:
@@ -84,7 +85,7 @@ def perform_action(svc, definition: Dict, params: Dict) -> Dict:
     state = params.get("state", None)
     kind = definition.get("kind")
     api_version = definition.get("apiVersion")
-    result = {}
+    result = {"changed": False, "result": {}}
 
     resource = svc.find_resource(kind, api_version, fail=True)
     existing = svc.retrieve(resource, definition)
--- a/plugins/module_utils/k8s/runner.py
+++ b/plugins/module_utils/k8s/runner.py
@@ -77,7 +77,7 @@ def perform_action(svc, definition: Dict, params: Dict) -> Dict:
     state = params.get("state", None)
     result = {}
 
-    resource = svc.find_resource(definition["kind"], definition["apiVersion"])
+    resource = svc.find_resource(definition.get("kind"), definition.get("apiVersion"))
     existing = svc.retrieve(resource, definition)
 
     if state == "absent":
@@ -91,7 +91,7 @@ def perform_action(svc, definition: Dict, params: Dict) -> Dict:
                 result["msg"] = (
                     "resource 'kind={kind},name={name},namespace={namespace}' "
                     "filtered by label_selectors.".format(
-                        kind=definition["kind"], name=origin_name, namespace=namespace,
+                        kind=definition.get("kind"), name=origin_name, namespace=namespace,
                     )
                 )
                 return result
--- a/plugins/module_utils/k8s/runner.py
+++ b/plugins/module_utils/k8s/runner.py
@@ -78,9 +78,15 @@ def perform_action(svc, definition: Dict, params: Dict) -> Dict:
     namespace = definition["metadata"].get("namespace")
     label_selectors = params.get("label_selectors")
     state = params.get("state", None)
+    kind = definition.get("kind")
+    api_version = definition.get("apiVersion")
     result = {}
 
-    resource = svc.find_resource(definition["kind"], definition["apiVersion"])
+    if kind and kind.endswith("List"):
+        resource = svc.find_resource(kind, api_version, fail=False)
+    else:
+        resource = svc.find_resource(kind, api_version, fail=True)
+
     existing = svc.retrieve(resource, definition)
 
     if state == "absent":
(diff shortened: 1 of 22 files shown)


plugins/modules/k8s.py
```diff
@@ -424,19 +399,20 @@ def main():
         ("template", "src"),
         ("name", "generate_name"),
     ]
-    module = AnsibleModule(
+
+    from ansible_collections.kubernetes.core.plugins.module_utils.k8s.core import (
+        AnsibleK8SModule,
+    )
+    from ansible_collections.kubernetes.core.plugins.module_utils.k8s.runner import (
```
Akasurde, line 406:
```
ModuleNotFoundError: No module named 'ansible_collections.kubernetes.core.plugins.module_utils.k8s.runner
```

alinabuzachis: This PR and the other migrating k8s_info required the runner to be merged first.

Akasurde: Yes, I saw that PR later. My bad.

alinabuzachis:
recheck

alinabuzachis:
recheck

alinabuzachis:
recheck

alinabuzachis:
recheck

alinabuzachis:
recheck

plugins/module_utils/k8s/runner.py
```diff
@@ -61,23 +63,30 @@ def run_module(module) -> None:
...
+    kind = definition.get("kind")
+    api_version = definition.get("apiVersion")
     result = {}
 
-    resource = svc.find_resource(definition)
+    if kind and kind.endswith("List"):
+        resource = svc.find_resource(kind, api_version, fail=False)
+    else:
+        resource = svc.find_resource(kind, api_version, fail=True)
+
```
gravesm, line 89: These changes shouldn't be necessary. If they are that means there's most likely a problem elsewhere.

plugins/module_utils/k8s/runner.py
```diff
@@ -61,23 +63,30 @@ def run_module(module) -> None:
...
-                module.fail_json(msg=e)
+                module.fail_json(msg=to_native(e))
+
         if module.warnings:
             result["warnings"] = module.warnings
 
         results.append(result)
 
-    module.exit_json(**results)
+    module.exit_json(result=results)
```
gravesm, line 73: I believe this is the reason for the current failure. We need to replicate the behavior here: https://github.com/ansible-collections/kubernetes.core/blob/ed33d0b56ed61857b370a7a87d340578ab531862/plugins/module_utils/common.py#L778-L781

plugins/module_utils/k8s/runner.py
alinabuzachis: If I remove these lines here and keep only `resource = svc.find_resource(kind, api_version)`, this tasks fails https://github.com/ansible-collections/kubernetes.core/blob/2.x-refactor/molecule/default/roles/helm/tasks/test_crds.yml#L30 since the result won't contain `'Failed to find exact match for example.com/v1.Foo'` (it won't go here https://github.com/ansible-collections/kubernetes.core/blob/2.x-refactor/plugins/module_utils/k8s/service.py#L94 because fail is False).

alinabuzachis:
recheck

alinabuzachis:
recheck

plugins/module_utils/k8s/runner.py
gravesm: Ah, so what needs to happen then is just keep the `svc.find_resource(..., fail=True)` line. The old code conflates resource retrieval with building the list of definitions. We have separated those two things out in the refactor, so we should never have a List kind resource here. And if we did, none of the rest of this function is going to work anyways, so we can go ahead and fail here.

alinabuzachis:
recheck

alinabuzachis:
recheck

alinabuzachis:
recheck

plugins/module_utils/k8s/service.py
```diff
@@ -514,7 +517,7 @@ def update(
             wait_condition = self.module.params["wait_condition"]
         results = {"changed": False, "result": {}}
 
-        if self.module.check_mode and not self.module.client.dry_run:
+        if self.module.check_mode and not self.client.dry_run:
             k8s_obj = dict_merge(existing.to_dict(), _encode_stringdata(definition))
         else:
             for merge_type in self.module.params.get("merge_type") or [
```
gravesm, line 523: The failure with the crd test is coming from here. We need to carry over the logic to test multiple merge types, since we are right now failing on the first try. So something like this in the `else` block:
```python
exception = None
for merge_type in self.module.params.get("merge_type") or ["strategic-merge", "merge",]:
    try:
        k8s_obj = self.patch_resource(...)
        exception = None
    except CoreException as e:
        exception = e
        continue
    break
if exception:
    raise exception
```

alinabuzachis: I removed the changed I make to the crd.yml file and added this block. But it keeps failing.

```
TASK [Patch custom resource definition] ****************************************
fatal: [localhost]: FAILED! => {"changed": false, "msg": "Failed to patch object: b'{\"kind\":\"Status\",\"apiVersion\":\"v1\",\"metadata\":{},\"status\":\"Failure\",\"message\":\"the body of the request was in an unknown format - accepted media types include: application/json-patch+json, application/merge-patch+json, application/apply-patch+yaml\",\"reason\":\"UnsupportedMediaType\",\"code\":415}\\n'"}
...ignoring

TASK [Assert that recreating crd is as expected] *******************************
    "assertion": "recreate_crd is not failed",
    "changed": false,
    "evaluated_to": false,
    "msg": "Assertion failed"
}
```

plugins/module_utils/k8s/service.py
```diff
@@ -514,16 +517,24 @@ def update(
...
                 "strategic-merge",
                 "merge",
             ]:
-                k8s_obj = self.patch_resource(
-                    resource, definition, name, namespace, merge_type=merge_type,
-                )
+                try:
+                    k8s_obj = self.patch_resource(
+                        resource, definition, name, namespace, merge_type=merge_type,
+                    )
```
gravesm, line 531: You need to add `exception = None` right here, otherwise the behavior is effectively the same as before this change.

alinabuzachis: Right! Thanks, missed that!

alinabuzachis:
recheck

alinabuzachis:
recheck

goneri:
recheck

alinabuzachis:
recheck

alinabuzachis:
recheck

alinabuzachis:
recheck

alinabuzachis:
recheck

PR #327, opened 2022-01-13
Refactor k8s_service to use new module_utils code

alinabuzachis:
##### SUMMARY

Refactor k8s_service to use new module_utils code

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

k8s_service

### DIFF
--- a/plugins/modules/k8s_service.py
+++ b/plugins/modules/k8s_service.py
@@ -144,11 +144,6 @@ result:
 
 import copy
 
-from collections import defaultdict
-
-from ansible_collections.kubernetes.core.plugins.module_utils.ansiblemodule import (
-    AnsibleModule,
-)
 from ansible_collections.kubernetes.core.plugins.module_utils.args_common import (
     AUTH_ARG_SPEC,
     COMMON_ARG_SPEC,
@@ -173,19 +168,6 @@ SERVICE_ARG_SPEC = {
 }
 
 
-def merge_dicts(x, y):
-    for k in set(x.keys()).union(y.keys()):
-        if k in x and k in y:
-            if isinstance(x[k], dict) and isinstance(y[k], dict):
-                yield (k, dict(merge_dicts(x[k], y[k])))
-            else:
-                yield (k, y[k])
-        elif k in x:
-            yield (k, x[k])
-        else:
-            yield (k, y[k])
-
-
 def argspec():
     """ argspec property builder """
     argument_spec = copy.deepcopy(AUTH_ARG_SPEC)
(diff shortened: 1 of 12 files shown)


alinabuzachis:
recheck

gravesm (commented):
Unfortunately, we can't use `run_module()` here as it's currently written. The `selector`, `type`, and `ports` parameters get lost. We should also force the `kind` value. I'll need to think about whether we want to change that function and how. It may be easiest for now to just replicate what you need from `run_module()` here. We don't need the continue on error functionality because this module currently only works with a single resource definition.

plugins/modules/k8s_service.py
```diff
@@ -195,51 +187,12 @@ def argspec():
...
-    selector = module.params.get("selector")
-    service_type = module.params.get("type")
-    ports = module.params.get("ports")
-
-    definition = defaultdict(defaultdict)
-
-    definition["kind"] = "Service"
-    definition["apiVersion"] = api_version
-
-    def_spec = definition["spec"]
```
alinabuzachis, line 212: @gravesm `def_spec` and `def_meta` seem never to be used anywhere.

gravesm: They are used, but in a way that's not obvious at first glance. Since those reference dictionaries, they are assigned by reference. Changes in those dictionaries are reflected in the `definition` which is used later.

alinabuzachis: Ah, ok! Thanks.

alinabuzachis:
recheck

plugins/modules/k8s_service.py
```diff
@@ -195,10 +212,29 @@ def argspec():
...
 
-def execute_module(module, k8s_ansible_mixin):
-    """ Module execution """
-    k8s_ansible_mixin.set_resource_definitions(module)
+def perform_action(svc, module, definition):
+    result = {}
+    module.warnings = []
+
+    if module.params["validate"] is not None:
+        module.warnings = validate(svc.client, module, definition)
```
gravesm, line 220: This module does not support the `validate` param. I was a bit surprised to see that this passed CI since it should throw an `AttributeError` exception. Upon further investigation, I found that there is only a single test for this module and it uses `ignore_errors: yes`. :scream:

plugins/modules/k8s_service.py
```diff
@@ -195,10 +212,29 @@ def argspec():
...
 
+    try:
+        result = perform_action(svc, definition, module.params)
+    except CoreException as e:
+        if module.warnings:
+            e["msg"] += "\n" + "\n    ".join(module.warnings)
+        module.fail_json(msg=e)
```
gravesm, line 227: I think `msg` needs to be a string.

plugins/modules/k8s_service.py
```diff
@@ -195,10 +212,29 @@ def argspec():
...
-    k8s_ansible_mixin.set_resource_definitions(module)
+def perform_action(svc, module, definition):
+    result = {}
+    module.warnings = []
+
+    if module.params["validate"] is not None:
+        module.warnings = validate(svc.client, module, definition)
 
+    try:
+        result = perform_action(svc, definition, module.params)
```
gravesm, line 223: This function is infinitely recursive as it just keeps calling itself. Did you mean to import the `perform_action()` function from `runner.py`?

gravesm (commented):
It may be worth adding one or two tests for this module at the very least just as a smoke test.

plugins/modules/k8s_service.py
alinabuzachis: 😭I wanted to replicate some things from the runner... but I didn't pay attention at the naming and the other params... my fault! 
I also noticed `k8s_service` is not fully covered with integration tests.

plugins/modules/k8s_service.py
alinabuzachis: It should call `perform_action()` from runner. But I used the wrong naming.

plugins/modules/k8s_service.py
alinabuzachis: Yes, it needs. Thanks!

plugins/modules/k8s_service.py
alinabuzachis: @gravesm Can we open some issues for the modules that are not fully covered by integration tests?

gravesm: Yes, definitely.

alinabuzachis:
recheck

alinabuzachis:
recheck

alinabuzachis:
recheck

Issue #291, opened 2021-11-23
turbo mode for lookup and inventory plugins


jobcespedes:
##### SUMMARY

Add turbo mode for lookup and inventory plugins
##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

plugin
##### ADDITIONAL INFORMATION

It will be used with Ansible Operator SDK to improve perfomance not only when using modules, but when using lookup and inventory plugins

N/A

gravesm:
@jobcespedes Thanks for the feature request. Adding turbo mode to the lookup plugins should be doable. I'm not sure I see how the inventory plugin would benefit from this as it's only called once at the start. I may be missing something, though?

jobcespedes:
Thanks for the reply and consideration. I mentioned the inventory plugin with one possible use case in mind: inventory plugin running inside an Ansible Operator. There, many reconciling loops may run frequently for several CR, using an inventory plugin. However, there might be some work to be done before:

- cloud.common does not yet offer an implementation for an inventory plugin
- operator SDK does not integrate closely with kubernetes.core inventory plugin

yurnov:
Is it still valid? As support of turbo mode was added to K8s lookup (as #335), and also `cloud.common` (that means turbo mode) is not compatible with the ansible-core>=2.19. and 2.18 will reach EOL in this month.

Suggest closing this issue.

att @beeankha @hdefazio

beeankha:
Thank you @jobcespedes and @yurnov , yes we can go ahead and close this issue.

PR #335, opened 2022-01-14
Enable turbo mode for k8s lookup plugin

abikouo:
##### SUMMARY

Enable Turbo mode for k8s lookup plugin
This resolves partially #291 

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

k8s lookup


plugins/lookup/k8s.py
```diff
@@ -183,15 +183,31 @@
...
-from ansible.plugins.lookup import LookupBase
+from ansible.module_utils.common.validation import check_type_bool
 
 from ansible_collections.kubernetes.core.plugins.module_utils.common import (
     K8sAnsibleMixin,
     get_api_client,
 )
 
+try:
+    enable_turbo_mode = check_type_bool(os.environ.get("ENABLE_TURBO_MODE"))
```
Akasurde, line 198: Could you please example in README.md or https://github.com/ansible-collections/kubernetes.core/blob/main/docs/ansible_turbo_mode.rst? So it would be easy for end-user to use.

abikouo: Thanks for the feedback.
README has been updated

Akasurde (approved):

PR #337, opened 2022-01-17
[backport/2.2] Update k8s_scale to use module_utils refactored code

alinabuzachis:
##### SUMMARY

Update `k8s_scale` to use `module_utils` refactored code

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

k8s_scale

### DIFF
--- a/plugins/modules/k8s_scale.py
+++ b/plugins/modules/k8s_scale.py
@@ -151,6 +151,22 @@ from ansible_collections.kubernetes.core.plugins.module_utils.args_common import
     RESOURCE_ARG_SPEC,
     NAME_ARG_SPEC,
 )
+from ansible_collections.kubernetes.core.plugins.module_utils.k8s.client import (
+    get_api_client,
+)
+from ansible_collections.kubernetes.core.plugins.module_utils.k8s.core import (
+    AnsibleK8SModule,
+)
+from ansible_collections.kubernetes.core.plugins.module_utils.k8s.exceptions import (
+    CoreException,
+    ResourceTimeout,
+)
+from ansible_collections.kubernetes.core.plugins.module_utils.k8s.service import (
+    K8sService,
+)
+from ansible_collections.kubernetes.core.plugins.module_utils.k8s.waiter import (
+    get_waiter,
+)
 
 
 SCALE_ARG_SPEC = {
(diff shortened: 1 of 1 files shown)


plugins/modules/k8s_scale.py
```diff
@@ -194,7 +202,10 @@ def execute_module(
...
         return_attributes["duration"] = 0
 
-    resource = k8s_ansible_mixin.find_resource(kind, api_version, fail=True)
+    svc = K8sService(client, module)
+    resource = svc.find_resource(
+        module.params.get("kind"), module.params.get("api_version"), fail=True
+    )
 
     from ansible_collections.kubernetes.core.plugins.module_utils.common import (
         NotFoundError,
```
gravesm, line 211: This should be removed and we should import this directly from kubernetes so we remove the dependency on common.py

plugins/modules/k8s_scale.py
```diff
@@ -277,14 +287,7 @@ def _continue_or_exit(warn):
                     result = resource.patch(existing.to_dict()).to_dict()
                 else:
                     result = scale(
```
gravesm, line 289: Wrap this scale() call in try/except looking for a CoreException and if it catches one call module.fail_json

alinabuzachis: Done!

alinabuzachis:
recheck

Akasurde:
recheck

alinabuzachis:
recheck

plugins/modules/k8s_scale.py
```diff
@@ -163,17 +187,9 @@
...
-
-    definition = k8s_ansible_mixin.resource_definitions[0]
-
-    name = definition["metadata"]["name"]
-    namespace = definition["metadata"].get("namespace")
-    api_version = definition["apiVersion"]
-    kind = definition["kind"]
+def execute_module(client, module):
+    namespace = module.params.get("namespace")
+    name = module.params.get("name")
```
gravesm, line 192: This needs to call `create_definitions()` from `resource.py` first, and then get `namespace`, `name`, `api_version` and `kind` from the returned definition.

gravesm (approved):

Issue #313, opened 2021-12-22
Missing flags on template module


Alfedi:
##### SUMMARY

Template command has a lot of flags not available in the `helm_template` module, for instance you cannot set a namespace to the generated template or a name.
##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

`helm_template`
##### ADDITIONAL INFORMATION

This will simply add at least the same parameters as the `helm` module. This way the module will be as useful as the helm template command. It would be something like this:

```yaml
- name: Generate template of Prometheus chart
  kubernetes.core.helm_template:
    name: test
    chart_ref: stable/prometheus
    release_namespace: monitoring
    create_namespace: true
    include_crds: true
    disable_hook: true
```

gravesm:
Thank you for the suggestion @Alfedi.

PR #348, opened 2022-01-21
helm_template: change order of values_release and values_files

cvstebut:
##### SUMMARY

This fix aligns precedence of release_values and values_files in kubernetes.core.helm_template with the one in kubernetes.core.helm. 
The values in release_values are now processed last, thus with highest precedence.
This allows overwriting of values in values_files with the values presented in release_values.

##### ISSUE TYPE

- Bugfix Pull Request

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/changelogs/fragments/348-helm_template-fix-precedence-of-release-values-over-values-files.yaml
+++ b/changelogs/fragments/348-helm_template-fix-precedence-of-release-values-over-values-files.yaml
@@ -1,0 +1,3 @@
+---
+bugfixes:
+  - helm_template - evaluate release_values after values_files, insuring highest precedence (now same behavior as in helm module). (https://github.com/ansible-collections/kubernetes.core/pull/348)
--- a/plugins/modules/helm_template.py
+++ b/plugins/modules/helm_template.py
@@ -152,15 +152,15 @@ def template(
     if output_dir:
         cmd += " --output-dir=" + output_dir
 
+    if values_files:
+        for values_file in values_files:
+            cmd += " -f=" + values_file
+
     if release_values:
         fd, path = tempfile.mkstemp(suffix=".yml")
         with open(path, "w") as yaml_file:
             yaml.dump(release_values, yaml_file, default_flow_style=False)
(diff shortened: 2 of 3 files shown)


Akasurde (commented):
Please create a [changelog fragment](https://docs.ansible.com/ansible/latest/community/development_process.html#creating-a-changelog-fragment). See [this fragment](https://github.com/ansible/ansible/blob/c2f528b7686c9573aa8493357219e0bbe4416dd3/changelogs/fragments/reboot-change-default-boot-command.yaml) as an example. Also, could you please add a test for this change?

Akasurde:
recheck

cvstebut:
Sorry it took a little while for the changelog fragment and test. For a newbie, changing the order of some lines is easy; the necessary test took a little bit more research.

I have included a unit test checking the order of the parameters in the "helm template ..." command string generated by the helm_template module.

In my view, the unit test should be enough in this case, as no module parameters were changed.

Not sure I have put it in the "correct" place, as it seems to be the first unit test for modules.

Also not sure, if it is ok to base the test on an "internal" function of the module; but this does allow to test the only relevant output of the module. An alternative would be to extract the generated command by defining a MockedModule including the run_command and extracting the command parameter. (similar to tests/unit/module_units/htest_helm.py)
Please advise.

I am thinking about adding other parameters to the module (e.g. release_name) (see #313) , but first I will  start looking into how to test module parameters and generated docs.

gravesm:
recheck

gravesm:
@cvstebut Thanks for the changelog and the tests. We would typically add a test for this in the integration test suite here: https://github.com/ansible-collections/kubernetes.core/tree/main/molecule/default/roles/helm/tasks, but the unit test works, as well.

cvstebut:
recheck

Akasurde:
recheck

Akasurde:
@cvstebut Due to a change in CI I am closing this PR and reopening it as https://github.com/ansible-collections/kubernetes.core/pull/373. Thanks for the contribution.

PR #352, opened 2022-01-31
k8s_secret: new module to create secrets

abikouo:
##### SUMMARY

create secrets on kubernetes analogous to ``kubectl create secret``

##### ISSUE TYPE

- New Module Pull Request

##### COMPONENT NAME

k8s_secret
##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/molecule/default/library/k8s_molecule_x509_cert.py
+++ b/molecule/default/library/k8s_molecule_x509_cert.py
@@ -1,0 +1,157 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+# Copyright: (c) 2022, Aubin Bikouo <@abikouo>
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+
+__metaclass__ = type
+
+
+DOCUMENTATION = r'''
+
+module: k8s_molecule_x509_cert.py
+
+short_description: Generate x509 certificate and private key pair.
+
+author:
+    - Aubin Bikouo (@abikouo)
+
+description:
+  - This module is used to generate x509 certificate and private key for molecule
+    testing for the kubernetes.core collection.
+  - This module is not supported outside of this context.
+
+options:
+  path:
+    description:
+    - The destination path where files will be created.
+    - The following files will be created certificate.pem and private_key.pem with the prefix on the file name.
+    type: path
+    required: yes
+  prefix:
+    description:
+    - generated files prefix.
+    type: str
+  x509_size:
+    description:
+    - certificate size.
+    type: int
+    choices:
+    - 2048
+    - 4096
+  x509_validity:
+    description:
+    - certificate validity in year.
+    type: int
+    default: 1
+
+requirements:
+  - python >= 3.6
+  - PyOpenSSL
+'''
+
+EXAMPLES = r'''
+- name: create x509 certificate into /tmp/certificates
+  k8s_molecule_x509_cert:
+    path: /tmp/certificates
+'''
+
+
+RETURN = r'''
+cert:
+  description:
+  - The base64 encoded value of the certificate generated.
+  returned: success
+  type: str
+key:
+  description:
+  - The base64 encoded value of the private key.
+  returned: success
+  type: str
+'''
+
+import traceback
+import socket
+import random
+import os
+import base64
+
+from ansible.module_utils.basic import AnsibleModule
+
+try:
+    from OpenSSL import crypto
+
+    HAS_OPENSSL_PY_LIBRARY = True
+except ImportError as e:
+    HAS_OPENSSL_PY_LIBRARY = False
+    pyopenssl_import_exception = e
+    IMPORT_OPENSSL_ERR = traceback.format_exc()
+
+
+def generate_self_signed_cert(key_size, validity):
+    pkey = crypto.PKey()
+    pkey.generate_key(crypto.TYPE_RSA, key_size)
+
+    x509 = crypto.X509()
+    subject = x509.get_subject()
+    subject.commonName = socket.gethostname()
+    x509.set_issuer(subject)
+    x509.gmtime_adj_notBefore(0)
+    x509.gmtime_adj_notAfter(validity * 365 * 24 * 60 * 60)
+    x509.set_pubkey(pkey)
+    x509.set_serial_number(random.randrange(100000))
+    x509.set_version(2)
+    host_extension = "DNS:%s" % socket.gethostname()
+    x509.add_extensions([
+        crypto.X509Extension(b"subjectAltName", False, host_extension.encode()),
+        crypto.X509Extension(b"basicConstraints", True, b"CA:false")]
+    )
+
+    x509.sign(pkey, 'SHA256')
+
+    return (crypto.dump_certificate(crypto.FILETYPE_PEM, x509),
+            crypto.dump_privatekey(crypto.FILETYPE_PEM, pkey))
+
+
+def execute_module(module):
+    if not HAS_OPENSSL_PY_LIBRARY:
+        module.fail_json(
+            msg="This module requires PyOpenSSL python library, install it using `pip install pyopenssl`"
+        )
+
+    params = dict(
+        key_size=module.params.get("x509_size"),
+        validity=module.params.get("x509_validity")
+    )
+    certificate, private_key = generate_self_signed_cert(**params)
+
+    path = module.params.get("path")
+    prefix = module.params.get("prefix")
+    with open(os.path.join(path, "%scertificate.pem" % prefix), "wb") as f:
+        f.write(certificate)
+
+    with open(os.path.join(path, "%sprivate_key.pem" % prefix), "wb") as f:
+        f.write(private_key)
+
+    module.exit_json(
+        changed=True,
+        cert=base64.b64encode(certificate).decode(),
+        key=base64.b64encode(private_key).decode(),
+    )
+
+
+def main():
+    argument_spec = dict(
+        path=dict(type='path', required=True),
+        prefix=dict(type='str', default=''),
+        x509_size=dict(type='int', choices=[2048, 4096]),
+        x509_validity=dict(type='int', default=1),
+    )
+    module = AnsibleModule(argument_spec=argument_spec)
+
+    execute_module(module)
+
+
+if __name__ == '__main__':
+    main()
--- a/plugins/action/k8s_info.py
+++ b/plugins/action/k8s_info.py
@@ -252,6 +252,17 @@ class ActionModule(ActionBase):
             new_module_args.pop("template")
         new_module_args["definition"] = result_template
 
+    def find_file(self, path):
+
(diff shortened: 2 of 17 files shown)


abikouo:
recheck

changelogs/fragments/352-k8s_secret-module.yml
```diff
@@ -0,0 +1,3 @@
+---
+major_changes:
+  - k8s_secret - New module for creating secrets, analogous to `kubectl create secret` (https://github.com/ansible-collections/community.kubernetes/pull/352).
```
Akasurde, line 3: I think, there is no need for this file for the new module.

molecule/default/library/k8s_molecule_x509_cert.py
```diff
@@ -0,0 +1,157 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+# Copyright: (c) 2021, Aubin Bikouo <@abikouo>
```
Akasurde, line 3:
```suggestion
# Copyright: (c) 2022, Aubin Bikouo <@abikouo>
```

plugins/action/k8s_info.py
```diff
@@ -263,16 +274,38 @@ def get_file_realpath(self, local_path):
...
+        if self._task.action not in (
+            "k8s_secret",
+            "kubernetes.core.k8s_secret",
+            "community.okd.k8s_secret",
+            "redhat.openshift.k8s_secret",
+            "community.kubernetes.k8s_secret",
+        ):
             raise AnsibleActionFail(
-                "%s does not exist in local filesystem" % local_path
+                "'tls_certificate' and 'from_path' are only supported parameters for the 'k8s_secret' module."
```
Akasurde, line 291:
```suggestion
                "'tls_certificate' and 'from_path' are only supported parameters for the 'kubernetes.core.k8s_secret' module."
```

plugins/modules/k8s_secret.py
```diff
@@ -0,0 +1,724 @@
...
+module: k8s_secret
+
+short_description: Create, update or delete a secret
+
+version_added: "2.3.0"
+
+author: Aubin Bikouo (@abikouo)
+
+description:
+  - Use Kubernetes Python SDK to create generic, tls or docker-registry secrets on Kubernetes.
```
Akasurde, line 23:
```suggestion
  - Use Kubernetes Python SDK to create C(generic), C(tls), or C(docker-registry) secrets on Kubernetes.
```

plugins/modules/k8s_secret.py
```diff
@@ -0,0 +1,724 @@
...
+  - Analogous to `kubectl create secret`.
+  - A generic secret can be created using a file, directory, or specified literal value.
+  - When creating a generic secret based on a file, the key will default to the basename of the file, and the value will default to
+    the file content. If the basename is an invalid key or you wish to chose your own, you may specify an alternate key.
+  - When creating a generic secret based on a directory, each file whose basename is a valid key in the directory will be packaged
+    into the secret. Any directory entries except regular files are ignored (e.g. subdirectories, symlinks, devices, pipes, etc).
```
Akasurde, line 29:
```suggestion
    into the secret. Any directory entries except regular files are ignored (for example, subdirectories, symlinks, devices, pipes, and so on).
```

plugins/modules/k8s_secret.py
```diff
@@ -0,0 +1,724 @@
...
+  - A TLS secret is created from a given public/private key pair.
+    The public/private key pair must exist before hand. The public key certificate must be .PEM encoded and match the given private key.
+  - A docker-registry secret is used to authenticate against Docker registries.
```
Akasurde, line 32:
```suggestion
  - A C(docker-registry) secret is used to authenticate against Docker registries.
```

plugins/modules/k8s_secret.py
```diff
@@ -0,0 +1,724 @@
...
+    - Append a hash of the secret to its name.
+    type: bool
+    default: false
+  type:
+    description:
+    - Specifies the type of Secret to create.
+    - When set to I(generic) one of C(literals) or C(from_path) should be set.
+    - When set to I(docker-registry) one of C(docker-registry-credentials), C(literals) or C(from_path) should be set.
+    - Ignored when C(state) is set to I(absent).
+    - PyOpenSSL python library is required to create tls secret.
```
Akasurde, line 61:
```suggestion
    - PyOpenSSL python library is required to create C(tls) secret.
```

plugins/modules/k8s_secret.py
```diff
@@ -0,0 +1,724 @@
...
+    required: true
+    type: str
+  namespace:
+    description:
+      - Use to specify a Secret object namespace.
+    required: true
+    type: str
+  docker_registry_credentials:
+    description:
+      - Provide details for docker registry authentication.
```
Akasurde, line 80:
```suggestion
      - Provide details for Docker registry authentication.
```

plugins/modules/k8s_secret.py
```diff
@@ -0,0 +1,724 @@
...
+    description:
+      - Specifies file or directory defining key and value to insert in secret.
+      - Key files can be specified using their file path, in which case a default name will be given to
+        them, or as dictionnary with a name and file path, in which case the given name will be used.
+      - Specifying a directory will iterate each named file in the directory that is a valid secret key.
+      - Ignored when C(type) is not set to I(generic).
+    type: raw
+  tls_certificate:
+    description:
+      - Provide tls certificate along with corresponding private key to create a new tls secret.
```
Akasurde, line 114:
```suggestion
      - Provide TLS certificate along with the corresponding private key to create a new C(tls) secret.
```

plugins/modules/k8s_secret.py
```diff
@@ -0,0 +1,724 @@
...
+    type: dict
+    suboptions:
+        certificate:
+            description:
+            - Path to PEM encoded public key certificate.
+            type: path
+            required: true
+        private_key:
+            description:
+            - Path to private key associated with given certificate.
```
Akasurde, line 126:
```suggestion
            - Path to the private key associated with the given certificate.
```

plugins/modules/k8s_secret.py
```diff
@@ -0,0 +1,724 @@
...
+            type: path
+            required: true
+  literals:
+    description:
+      - dictionnary of key and literal value to insert in secret data.
```
Akasurde, line 131:
```suggestion
      - Dictionary of key and literal value to insert in secret data.
```

plugins/modules/k8s_secret.py
```diff
@@ -0,0 +1,724 @@
...
+    description:
+      - Specifies the type of generic secret to create.
+      - When set to I(kubernetes.io/service-account-token) you need to ensure that the C(annotations) contains
+        key 'kubernetes.io/service-account.name' set to an existing service account name. The module may report that
+        the secret has been created when the service account does not exist, but it is not.
+      - Ignores when C(type) is not set to I(generic).
+    type: str
+    default: Opaque
+    choices:
+    - Opaque
```
Akasurde, line 144:
```suggestion
    - opaque
```
Can we make this all lowercase just like other options?

plugins/modules/k8s_secret.py
```diff
@@ -0,0 +1,724 @@
...
+    - kubernetes.io/service-account-token
+    - kubernetes.io/dockerconfigjson
+    - kubernetes.io/basic-auth
+    - kubernetes.io/ssh-auth
+    - bootstrap.kubernetes.io/token
+  labels:
+    description:
+    - dictionnary of string keys and values used as labels for the secret to be created.
```
Akasurde, line 152:
```suggestion
    - A dictionary of string keys and values is used as labels for the secret to be created.
```

plugins/modules/k8s_secret.py
```diff
@@ -0,0 +1,724 @@
...
+    - see U(http://kubernetes.io/docs/user-guide/labels).
+    - Ignored if C(state) is set to I(absent).
+    type: dict
+  annotations:
+    description:
+    - dictionnary of string keys and values used as annotations for the secret to be created.
```
Akasurde, line 158:
```suggestion
    - A dictionary of string keys and values is used as annotations for the secret to be created.
```

plugins/modules/k8s_secret.py
```diff
@@ -0,0 +1,724 @@
...
+- name: Create a new TLS secret
+  kubernetes.core.k8s_secret:
+    name: my-tls-secret
+    namespace: my-secret-namespace
+    type: tls
+    tls_certificate:
+        certificate: path/to/tls.key
+        private_key: path/to/tls.cert
+
+- name: Create a  docker configuration secret
```
Akasurde, line 211:
```suggestion
- name: Create a Docker configuration secret
```

Akasurde (changes requested):

plugins/modules/k8s_secret.py
abikouo: ok I will do

changelogs/fragments/352-k8s_secret-module.yml
abikouo: ok thanks will be removed

plugins/modules/k8s_secret.py
```diff
@@ -0,0 +1,724 @@
...
+            description:
+            - Email for Docker registry.
+            type: str
+  from_path:
+    description:
+      - Specifies file or directory defining key and value to insert in secret.
+      - Key files can be specified using their file path, in which case a default name will be given to
+        them, or as dictionnary with a name and file path, in which case the given name will be used.
+      - Specifying a directory will iterate each named file in the directory that is a valid secret key.
+      - Ignored when C(type) is not set to I(generic).
```
Akasurde, line 110:
```suggestion
    description:
      - Specifies a file or directory defining key and value to insert in secret.
      - Key files can be specified using their file path, in which case a default name will be given to
        them, or a dictionary with a name and file path, in which case the given name will be used.
      - Specifying a directory will iterate each named file in the directory that is a valid secret key.
      - Ignored when C(type) is not set to I(generic).
```

Akasurde (approved):

abikouo:
recheck

abikouo:
recheck

gravesm:
Thanks for your work on this. I'm not sure if this is something we want to merge at this time, though. We have generally tried to avoid creating modules for specific kubernetes resources.

abikouo:
> Thanks for your work on this. I'm not sure if this is something we want to merge at this time, though. We have generally tried to avoid creating modules for specific Kubernetes resources.

I am getting the point but this is currently done for the service with the module ``k8s_service`` and today users are forced to use ``kubectl`` executable if they want to create a secret, so I think this is definitely something which will be helpful for this collection.

gravesm:
`k8s_service` is an outlier and something I would actually prefer to try and eventually remove. You can still create a secret with the `k8s` module, and this would be easy to do if your secrets were stored in vault or some other secrets management solution, which seems like the more likely scenario. Part of the problem with this approach is that we're creating two different ways of managing the same k8s resource, making it more likely to have divergent behavior.

PR #354, opened 2022-01-31
helm_repository: Added support for common options

Akasurde:
Depends-On: https://github.com/ansible-collections/kubernetes.core/pull/366

##### SUMMARY

Added support for ``host``, ``api_key``, ``ca_cert``,
and ``validate_certs`` in helm_repository module.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
changelogs/fragments/helm_repository.yml
plugins/modules/helm_repository.py

### DIFF
--- a/plugins/modules/helm_repository.py
+++ b/plugins/modules/helm_repository.py
@@ -72,6 +72,31 @@ options:
     default: false
     type: bool
     version_added: 2.3.0
+  host:
+    description:
+    - Provide a URL for accessing the API. Can also be specified via C(K8S_AUTH_HOST) environment variable.
+    type: str
+    version_added: "2.3.0"
+  api_key:
+    description:
+    - Token used to authenticate with the API. Can also be specified via C(K8S_AUTH_API_KEY) environment variable.
+    type: str
+    version_added: "2.3.0"
+  validate_certs:
+    description:
+    - Whether or not to verify the API server's SSL certificates. Can also be specified via C(K8S_AUTH_VERIFY_SSL)
+      environment variable.
+    type: bool
+    aliases: [ verify_ssl ]
+    default: True
+    version_added: "2.3.0"
+  ca_cert:
+    description:
+    - Path to a CA certificate used to authenticate with the API. The full certificate chain must be provided to
+      avoid certificate validation errors. Can also be specified via C(K8S_AUTH_SSL_CA_CERT) environment variable.
+    type: path
+    aliases: [ ssl_ca_cert ]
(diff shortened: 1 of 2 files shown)


plugins/modules/helm_repository.py
```diff
@@ -72,6 +72,8 @@
     default: false
     type: bool
     version_added: 2.3.0
+extends_documentation_fragment:
```
abikouo, line 75: This is a bit confusing because the related doc mentions the parameters ``api_key``, ``validate_certs``, ``ca_cert`` added in version ``1.2.0``, but with this modification should be something like ``2.2.4``

plugins/modules/helm_repository.py
```diff
@@ -204,6 +206,22 @@ def main():
...
+            ),
+            validate_certs=dict(
+                type="bool",
+                default=True,
+                aliases=["verify_ssl"],
+                fallback=(env_fallback, ["K8S_AUTH_VERIFY_SSL"]),
+            ),
+            api_key=dict(
+                type="str", no_log=True, fallback=(env_fallback, ["K8S_AUTH_API_KEY"])
+            ),
```
abikouo, line 224: you should also update the function ``install_repository`` above to use the new parameters in the helm command.
Based on the ``helm repo add`` help this should be ``--kube-apiserver``, etc..

abikouo (changes requested):

plugins/modules/helm_repository.py
Akasurde: I will update it.

plugins/modules/helm_repository.py
```diff
@@ -72,6 +72,31 @@
     default: false
     type: bool
     version_added: 2.3.0
+  host:
+    description:
+    - Provide a URL for accessing the API. Can also be specified via C(K8S_AUTH_HOST) environment variable.
+    type: str
+    version_added: "2.2.4"
```
gravesm, line 79: These should all be 2.3.0. This is a new feature.

abikouo (approved):

plugins/modules/helm_repository.py
Akasurde: I forgot to commit 😉 lol

gravesm (approved):

goneri:
recheck

goneri:
recheck

Akasurde:
recheck

Akasurde:
recheck

goneri:
recheck

PR #357, opened 2022-02-01
Added missing kubeconfig parameter to helm_cmd_common

suppaduppax:
Depends-On: https://github.com/ansible-collections/kubernetes.core/pull/366

--kubeconfig kubeconfig was not being appended to the helm command. 
This is a propsed fix, although I'm not sure if the correct location for it would be in helm_cmd_common. release_status() requires it right away so I added it to help_cmd_common rather than adding a kubeconfig parameter to each function requiring kubeconfig

##### ISSUE TYPE

- Bugfix Pull Request
- Docs Pull Request
- Feature Pull Request
- New Module Pull Request

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/plugins/modules/helm.py
+++ b/plugins/modules/helm.py
@@ -668,7 +668,8 @@ def main():
     skip_crds = module.params.get("skip_crds")
     history_max = module.params.get("history_max")
     timeout = module.params.get("timeout")
-
+    kubeconfig = module.params.get("kubeconfig")
+    
     if bin_path is not None:
         helm_cmd_common = bin_path
     else:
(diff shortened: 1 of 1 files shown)


plugins/modules/helm.py
```diff
@@ -677,6 +678,9 @@ def main():
     if update_repo_cache:
         run_repo_update(module, helm_cmd_common)
 
+    if kubeconfig:
```
Akasurde, line 681: @suppaduppax Thanks for the contribution. This is already handled in [here](https://github.com/ansible-collections/kubernetes.core/blob/b54e9ef4ef36e8da5a9a30b0c2d39232c2773aa8/plugins/module_utils/helm.py#L48). So I think this check is irrelevant.

Akasurde (changes requested):

goneri:
recheck

goneri:
recheck

goneri:
recheck

goneri:
recheck

goneri:
recheck

goneri:
recheck

gravesm:
I'm going to close this one as the module already correctly uses the specified kubeconfig. It is temporarily set on the environment. If you find that it is not doing this, feel free to open an issue with steps to reproduce the problem.

PR #366, opened 2022-02-04
k8scopy: rely on existing kubectl binary

Akasurde:
Depends-On: https://github.com/ansible-collections/kubernetes.core/pull/367

##### SUMMARY

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
molecule/default/roles/k8scopy/defaults/main.yml
molecule/default/roles/k8scopy/tasks/main.yml
molecule/default/roles/k8scopy/tasks/test_copy_directory.yml
molecule/default/roles/k8scopy/tasks/test_copy_file.yml
molecule/default/roles/k8scopy/tasks/test_copy_large_file.yml
molecule/default/roles/k8scopy/tasks/test_multi_container_pod.yml

### DIFF
--- a/molecule/default/roles/k8scopy/defaults/main.yml
+++ b/molecule/default/roles/k8scopy/defaults/main.yml
@@ -11,5 +11,3 @@ pod_with_two_container:
   container:
   - container-10
   - container-11
-
-kubectl_path: /tmp/kubectl
--- a/molecule/default/roles/k8scopy/tasks/main.yml
+++ b/molecule/default/roles/k8scopy/tasks/main.yml
@@ -1,15 +1,5 @@
 ---
 - block:
-    - name: Download kubeclt executable used to compare results
-      get_url:
-        url: https://dl.k8s.io/release/v1.21.3/bin/linux/amd64/kubectl
-        dest: "{{ kubectl_path }}"
-
-    - name: make kubectl executable
-      ansible.builtin.file:
-        path: "{{ kubectl_path }}"
-        mode: "+x"
-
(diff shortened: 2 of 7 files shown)


abikouo:
I would rather prefer keeping the existing and just check if there is a version installed

abikouo:
I would rather prefer doing something like 

```
- name: check if script is existing
  command: "command -v kubectl"
  register: result
  ignore_errors: true

- name: set path when existing
  set_fact:
     kubectl_path: "{{ result.stdout }}"
  when: result.rc == 0

- block: 
   (...) Download kubectl here

  when: result.rc != 0
```

This will allow running the test on environemnts where the kubectl is not installed.

goneri:
recheck

goneri:
> I would rather prefer doing something like
> 
> ```
> - name: check if script is existing
>   command: "command -v kubectl"
>   register: result
>   ignore_errors: true
> 
> - name: set path when existing
>   set_fact:
>      kubectl_path: "{{ result.stdout }}"
>   when: result.rc == 0
> 
> - block: 
>    (...) Download kubectl here
> 
>   when: result.rc != 0
> ```
> 
> This will allow running the test on environemnts where the kubectl is not installed.

I like @Akasurde approach because it simplify the structure of the tests and the new requirement is really reasonable IMO.

goneri:
For the record, I also pushed a similar change here: https://github.com/ansible/ansible-zuul-jobs/commit/ebaf966fe782f0f47099cb84229f740e31b4421a

Akasurde:
mergeit

PR #367, opened 2022-02-04
remove the .zuul.d directory

goneri:
We now avoid as much as possible the local Zuul configuration because
it's easy to break them inadvertently.


goneri:
recheck

goneri:
recheck

PR #375, opened 2022-02-09
Move diff and wait to perform_action

gravesm:
##### SUMMARY

This primarily moves the diff and wait logic from the various service
methods to perform_action to eliminate code duplication. I also moved
the diff_objects function out of the service object and moved most of
the find_resource logic to a new resource client method. We ended up
with several modules creating a service object just to use one of these
methods, so it seemed to make sense to make these more accessible.

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/plugins/module_utils/k8s/client.py
+++ b/plugins/module_utils/k8s/client.py
@@ -27,9 +27,15 @@ except ImportError:
 
 try:
     import kubernetes
+    from kubernetes.dynamic.exceptions import (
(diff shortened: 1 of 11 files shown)


gravesm:
recheck

gravesm:
recheck

gravesm:
recheck

alinabuzachis (approved):
Nice!

PR #390, opened 2022-02-22
k8copy: relax an error msg check

goneri:
The msg string should actually be prefix with `Failed to copy object due to: `.

See: https://ansible.softwarefactory-project.io/zuul/build/50a9b85891c545149ab2ae35eaf73d22


Akasurde:
recheck

Akasurde:
File moved to `tests/`. I will revisit if required.

Issue #30, opened 2020-10-01
Helm Post Rendering support


gaborho:
##### SUMMARY
Helm Post Rendering support

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
community.kubernetes.helm

##### ADDITIONAL INFORMATION
Post rendering gives chart installers the ability to manually manipulate, configure, and/or validate rendered manifests before they are installed by Helm. It would be great to support this in the helm module. Use case would be to hand over a post renderer script (bash, python) what is execuring kustomize customization. More information: https://helm.sh/docs/topics/advanced/#post-rendering

Prometej:
I am also interested in this feature. I would like to run kustomize in post-renderer. This is very useful if you need to make updates (usually security issues that are not covered in templating) on community charts without modifying them.

The links you provided are currently broken could you please update them?

tima:
@Prometej I think ansible-collections/community.kubernetes#367 became #39 when the repo was migrated.

akamac:
@Akasurde @tima it is a different feature. Post-rendering is not limited to kustomize, but rather can call any executable. This is a matter of exposing the `--post-renderer` flag in module args.

Akasurde:
@akamac Would you be interested in opening a PR with your findings? Thanks in advance.

PR #403, opened 2022-03-11
Added support for Helm post-renderer

mnaser:
##### SUMMARY

Add support for Helm post renderer

Fixes: #30 

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

`helm`

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/plugins/modules/helm.py
+++ b/plugins/modules/helm.py
@@ -146,7 +146,7 @@ options:
     description:
       - Path to an executable to be used for post rendering.
     type: str
-    version_added: "2.3.0"
+    version_added: "2.4.0"
   replace:
     description:
       - Reuse the given name, only if that name is a deleted release which remains in the history.
--- a/plugins/modules/helm.py
+++ b/plugins/modules/helm.py
@@ -142,6 +142,11 @@ options:
     type: bool
     default: False
     version_added: "0.11.1"
+  post_renderer:
+    description:
+      - Path to an executable to be used for post rendering.
+    type: str
+    version_added: "2.3.0"
   replace:
     description:
       - Reuse the given name, only if that name is a deleted release which remains in the history.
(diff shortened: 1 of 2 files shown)


plugins/modules/helm.py
```diff
@@ -453,6 +459,9 @@ def deploy(
             yaml.dump(release_values, yaml_file, default_flow_style=False)
         deploy_command += " -f=" + path
 
+    if post_renderer:
+        deploy_command = " --post-renderer=" + post_renderer
```
Akasurde, line 463: @mnaser Do you think we should add a logic to see if `post_renderer` binary exists or not? Or helm command takes care of that?

plugins/modules/helm.py
```diff
@@ -142,6 +142,11 @@
     type: bool
     default: False
     version_added: "0.11.1"
+  post_renderer:
+    description:
+      - Path to an executable to be used for post rendering.
+    type: str
+    version_added: "2.3.0"
```
Akasurde, line 149:
```suggestion
    version_added: "2.4.0"
```

Akasurde (changes requested):

plugins/modules/helm.py
mnaser: It's best to let Helm take care of it:

```
❯ helm install --post-renderer=foo
Error: invalid argument "foo" for "--post-renderer" flag: unable to find binary at foo: exec: "foo": executable file not found in $PATH
```

Akasurde: Great.

Akasurde:
recheck

Akasurde (approved):

Akasurde:
@gravesm Can you please take a look? Thanks

gravesm (approved):
@mnaser Thank you for the PR!

Issue #355, opened 2022-01-31
Ability to run `helm pull` as a task


jack1902:
##### SUMMARY

Basically, i'd like the ability to run ansible to download a bunch of helm charts for use in a separate ansible playbook that wouldn't have any access to the internet.

`helm_pull` would download all packages locally for use by another ansible-playbook which assumes all charts already exist at a given path.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

`kubernetes.core.helm_pull` or `kubernetes.core.helm_repository` could be suitable for this.

##### ADDITIONAL INFORMATION

Currently, the `helm` tasks assume that the host running the helm commands has access to the internet but this is not always the case.

```yaml
- name: Add linkerd chart repo
  kubernetes.core.helm_repository:
    name: linkerd
    repo_url: "https://helm.linkerd.io/stable"

- name: download linkerd chart locally
  kubernetes.core.helm_pull:
    name: linkerd/linkerd2
    # defaults to latest version
    version: VERSION_HERE
    # below would then contain a subdirectory with `linkerd.tgz` chart present or could be dumped out to the `files` directory
    dest: "/tmp/helm_charts"
```

rieger07:
+1 for this, would get rid of some shell commands in my playbooks

PR #410, opened 2022-03-17
helm - new module to perform helm pull

abikouo:
Depends-On: https://github.com/ansible/ansible-zuul-jobs/pull/1586

##### SUMMARY

#355 
new module to manage chart downloading ``helm pull``

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

helm_pull

### DIFF
--- a/plugins/modules/helm_pull.py
+++ b/plugins/modules/helm_pull.py
@@ -108,6 +108,11 @@ options:
     - Identify HTTPS client using this SSL key file
     - Requires helm >= 3.1.0.
     type: path
+  binary_path:
+    description:
+      - The path of a helm binary to use.
+    required: false
+    type: path
 """
 
 EXAMPLES = r"""
@@ -164,7 +169,7 @@ rc:
   sample: 1
 """
 
-from ansible.module_utils.basic import AnsibleModule, env_fallback
+from ansible.module_utils.basic import AnsibleModule
 from ansible_collections.kubernetes.core.plugins.module_utils.helm import (
     run_helm,
     get_helm_version,
@@ -194,6 +199,7 @@ def main():
         chart_ca_cert=dict(type="path"),
         chart_ssl_cert_file=dict(type="path"),
         chart_ssl_key_file=dict(type="path"),
+        binary_path=dict(type="path"),
     )
     module = AnsibleModule(
         argument_spec=argspec,
--- a/plugins/modules/helm_pull.py
+++ b/plugins/modules/helm_pull.py
@@ -108,9 +108,6 @@ options:
     - Identify HTTPS client using this SSL key file
     - Requires helm >= 3.1.0.
     type: path
-
-extends_documentation_fragment:
-  - kubernetes.core.helm_common_options
 """
 
 EXAMPLES = r"""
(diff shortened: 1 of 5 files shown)


goneri:
recheck

gravesm (commented):
A few small things, but otherwise looks good.

plugins/modules/helm_pull.py
```diff
@@ -0,0 +1,330 @@
...
+
+    helm_pull_opt_versionning = dict(
+        skip_tls_certs_check="3.3.0",
+        chart_ca_cert="3.1.0",
+        chart_ssl_cert_file="3.1.0",
+        chart_ssl_key_file="3.1.0",
+    )
+
+    helm_pull_opts = []
+    for k, v in iteritems(helm_pull_mapping):
```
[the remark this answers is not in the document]
abikouo: will do, thanks

plugins/modules/helm_pull.py
```diff
@@ -0,0 +1,330 @@
...
+        if module.params.get(k) is not None and module.params.get(k) != argspec.get(
```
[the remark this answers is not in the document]
abikouo: some parameters are boolean, so this means the corresponding flag should be passed to the command line only if the ansible task is called with a value different from the default value.

plugins/modules/helm_pull.py
```diff
@@ -0,0 +1,330 @@
...
+                opt_req_version
+            ):
+                module.fail_json(
+                    msg="Parameter {0} requires helm >= {1}, current version is {2}".format(
+                        k, opt_req_version, helm_version
+                    )
+                )
+            argtype = argspec.get(k).get("type", "str")
+            if argtype == "bool":
+                if module.params.get(k) != argspec.get(k).get("default", False):
```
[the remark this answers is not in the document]
abikouo: This comparison is useless as we already did the same above

plugins/modules/helm_pull.py
```diff
@@ -0,0 +1,335 @@
...
+        verify_chart=dict(key="verify"),
+        provenance=dict(key="prov"),
+        pass_credentials=dict(key="pass-credentials"),
+        skip_tls_certs_check=dict(key="insecure-skip-tls-verify"),
+        chart_devel=dict(key="devel"),
+        untar_chart=dict(key="untar"),
+    )
+
+    for k, v in helm_flag_args.items():
+        if module.params.get(k) and module.params.get(k) != v.get("default", False):
```
gravesm, line 302: This is still broken. If the default value were True and the user passed in True, or just left it unset, it's going to not add the flag. Can we just remove any comparison to a default value at all? I don't think there's any need for it, and it is adding unnecessary complexity. These are boolean values that control whether or not a flag is added to the command line. All we care about is whether the value we get in module.params is True or False. If it's True we set the flag, if it's False we don't.

abikouo: updated

plugins/modules/helm_pull.py
```diff
@@ -0,0 +1,335 @@
...
+    - Requires helm >= 3.1.0.
+    type: path
+  chart_ssl_key_file:
+    description:
+    - Identify HTTPS client using this SSL key file
+    - Requires helm >= 3.1.0.
+    type: path
+
+extends_documentation_fragment:
+  - kubernetes.core.helm_common_options
```
gravesm, line 113: Does this actually use any of these options? Unless I'm missing something, it doesn't seem like it's necessary.

gravesm (commented):
Sorry for taking so long to get to this. I think the module params just need to be cleaned up.

plugins/modules/helm_pull.py
abikouo: Removed.

gravesm (approved):

Issue #417, opened 2022-03-22
Clarify `return_code` deprecation warning


tima:
##### SUMMARY

As reported in #413, the deprecation warning emitted regarding the eventual removal of the `return_code` in favor of `rc` could be clearer to avoid confusion.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

- `kubernetes.core.k8s_exec`

##### ADDITIONAL INFORMATION

Here is a suggested rewording based on [a comment](https://github.com/ansible-collections/kubernetes.core/issues/413#issuecomment-1074277309) from @erikgb:

>  [DEPRECATION WARNING]: The 'return_code' return key is being renamed to 'rc'. 
Both keys are being returned for now to allow users to migrate their automation. 
The 'return_code' key will be removed from kubernetes.core in version  4.0.0. 
Deprecation warnings can be disabled by setting  deprecation_warnings=False 
in ansible.cfg.

PR #419, opened 2022-03-23
k8s_exec: Update deprecation warning

Akasurde:
##### SUMMARY

return_code is deprecated in favor of rc, update the
deprecation warning to tell user about the reason behind
this.

Fixes: #417

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
changelogs/fragments/417_deprecation.yml
plugins/modules/k8s_exec.py

### DIFF
--- a/plugins/modules/k8s_exec.py
+++ b/plugins/modules/k8s_exec.py
@@ -212,7 +212,7 @@ def execute_module(module, k8s_ansible_mixin):
         rc = int(err["details"]["causes"][0]["message"])
 
     module.deprecate(
-        "The 'return_code' return key is deprecated. Please use 'rc' instead. "
+        "The 'return_code' return key is being renamed to 'rc'. "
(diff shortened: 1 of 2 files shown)


plugins/modules/k8s_exec.py
```diff
@@ -212,7 +212,8 @@ def execute_module(module, k8s_ansible_mixin):
         rc = int(err["details"]["causes"][0]["message"])
 
     module.deprecate(
-        "The 'return_code' return key is deprecated. Please use 'rc' instead.",
+        "The 'return_code' return key is deprecated. Please use 'rc' instead. "
```
gravesm, line 215:
```suggestion
        "The 'return_code' return key is being renamed to 'rc'. "
```
I think the fact that we specifically tell the user to do something that they are already doing is partly why this is confusing.

gravesm (approved):

PR #436, opened 2022-04-20
Adding previous container log support

jeason81:
Signed-off-by: Joshua Eason <josh.eason@anchore.com>

##### SUMMARY
Adds support for the `previous` parameter in `kubectl logs`. This allows for the retrieval of the previously terminated containers logs which is useful for troubleshooting.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
k8s_log

##### ADDITIONAL INFORMATION
Adds the `previous` parameter (bool) to `k8s_log` module. This matches the documentation for `kubectl logs --previous` parameter. This parameter allows for retrieving the previously terminated containers logs.

Output of the module is identical with the exception being the logs returned are from the previously terminated container.

### DIFF
--- a/plugins/modules/k8s_log.py
+++ b/plugins/modules/k8s_log.py
@@ -60,6 +60,11 @@ options:
     required: no
     type: str
     version_added: '2.2.0'
+  previous:
+    description:
+    - If true, print the logs for the previous instance of the container in a pod if it exists. 
+    required: no
+    type: bool
 
 requirements:
   - "python >= 3.6"
@@ -138,6 +143,7 @@ def argspec():
             container=dict(),
             since_seconds=dict(),
             label_selectors=dict(type="list", elements="str", default=[]),
+            previous=dict(type="bool", default=False),
         )
     )
     return args
@@ -183,6 +189,11 @@ def execute_module(module, k8s_ansible_mixin):
     if module.params.get("since_seconds"):
         kwargs.setdefault("query_params", {}).update(
             {"sinceSeconds": module.params["since_seconds"]}
+        )
+
+    if module.params.get("previous"):
+        kwargs.setdefault("query_params", {}).update(
+            {"previous": module.params["previous"]}
         )
 
     log = serialize_log(


abikouo:
@jeason81 thanks for contributing to this repository. For new features, you need to add a changelog to your pull request.
as possible try please to add some integrated tests.

gravesm (commented):
Other than the changelog (you can find instructions [here](https://docs.ansible.com/ansible/latest/community/development_process.html#creating-a-changelog-fragment)), could you add a default of false to the new parameter? I know it's not strictly necessary, but it will make the docs a bit clearer.

plugins/modules/k8s_log.py
```diff
@@ -60,6 +60,11 @@
     required: no
     type: str
     version_added: '2.2.0'
+  previous:
+    description:
+    - If true, print the logs for the previous instance of the container in a pod if it exists.
+    required: no
+    type: bool
```
gravesm, line 67: Just realized this also needs a `version_added: '2.4.0'`.

jeason81:
I set the default to False and have updated the changelog. I'm not entirely sure the changelog is how you wanted it though. If not, let me know and I can correct. 

As for the testing, I'm a little curious how that could be supported in this instance. For this to work, we would need one of the containers to crash and restart within the test environment thus presenting the previous logs to the API. I'm not sure how we could force such a situation cleanly though I'm open to suggestions.

Thanks for all the assistance.

plugins/modules/k8s_log.py
```diff
@@ -60,6 +60,13 @@
     required: no
     type: str
     version_added: '2.2.0'
+  previous:
+    description:
+    - If true, print the logs for the previous instance of the container in a pod if it exists.
```
Akasurde, line 65:
```suggestion
    - If C(true), print the logs for the previous instance of the container in a pod if it exists.
```

plugins/modules/k8s_log.py
jeason81: This has been added.

gravesm (approved):
Thanks!

PR #455, opened 2022-05-09
Bring docs changes over from latest release

gravesm:
##### SUMMARY

Brings docs generation changes over from 2.3.1 release.

##### ISSUE TYPE

- Docs Pull Request

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -5,6 +5,66 @@ Kubernetes Collection Release Notes
 .. contents:: Topics
 
 
+v2.3.1
+======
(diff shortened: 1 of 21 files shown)


alinabuzachis:
> Unfortunately, that error seems to be the result of a change in Ansible 2.13, but I'm having a hard time figuring out exactly what the cause is or how to fix it.

I cannot reproduce it locally with latests versions to ansible-core and ansible.

goneri:
recheck

PR #478, opened 2022-06-15
Support resource definition using manifest URL

### LINKED ISSUE #451: Support kubectl apply from URL
##### SUMMARY

It should be possible to specify a manifest URL in the `kubernetes.core.k8s` module that gets passed on to `kubectl apply -f <URL>` under the hood.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
kubernetes.core.k8s

##### ADDITIONAL INFORMATION

Due to firewall issues or nasty HTTPS-breaking sniffing software (e.g. Netskope) on an administrator's laptop, it may be desirable to specify the manifest URL in the module configuration, so that the target machine downloads the manifest and applies it.

`get_url` technically works, but needs an extra step in the playbook and a place to store the temporary file.

`resource_definition: "{{ lookup('url', 'https://internal.corp/foo/bar.yml') }}"` has the problem that the case of "the only network allowed to access the manifest is a specific VPC" is not supported (as lookup will only run on the administrator's laptop).

```yaml
    kubernetes.core.k8s:
      state: present
      manifest_url: "https://docs.projectcalico.org/archive/v3.21/manifests/calicoctl.yaml"
```

abikouo:
##### SUMMARY

Closes #451 

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

``k8s``
``k8s_scale``
``k8s_service``


plugins/module_utils/common.py
```diff
@@ -520,8 +521,15 @@ def remove_aliases(self):
...
 
-    def load_resource_definitions(self, src):
+    def load_resource_definitions(self, src, module=None):
         """Load the requested src path"""
+        if module and (
+            src.startswith("https://")
+            or src.startswith("http://")
+            or src.startswith("ftp://")
+        ):
+            src = fetch_file(module, src)
```
gravesm, line 531: I think using `fetch_file()` here is potentially a problem. This will use the same `validate_certs` value that's used for connecting to the cluster, and we have no control over whether or not it does this. I would suggest that we use `open_url()` instead, or just use `requests` since we are already dependent on this anyways.

plugins/doc_fragments/k8s_resource_options.py
```diff
@@ -29,6 +29,7 @@ class ModuleDocFragment(object):
     - Reads from the local file system. To read from the Ansible controller's file system, including vaulted files, use the file lookup
       plugin or template lookup plugin, combined with the from_yaml filter, and pass the result to
       I(resource_definition). See Examples below.
+    - URL to files to use to create the resource. Added in version 2.4.0.
```
Akasurde, line 32:
```suggestion
    - The URL to manifest files that can be used to create the resource. Added in version 2.4.0.
```

plugins/module_utils/k8s/service.py
```diff
@@ -435,19 +434,21 @@ def delete(
         resource: Resource,
         definition: Dict,
         existing: Optional[ResourceInstance] = None,
-    ) -> Dict:
+    ) -> Tuple[Dict, bool]:
```
gravesm, line 437: We can't change the return type of this function because it would break https://github.com/ansible-collections/kubernetes.core/blob/9f51fc0ef0319e84b54e8cbab28508307a11c34c/plugins/module_utils/k8s/runner.py#L98. The tests are likely passing because we aren't doing a deep analysis of the returned object. There should be no need for these changes, however (see comment below).

plugins/modules/k8s_service.py
```diff
@@ -213,28 +214,35 @@ def argspec():
 
 def perform_action(svc, resource, definition, params):
```
gravesm, line 215: It looks to me like we can just replace this function with a call to `module_utils.k8s.runner.perform_action()`.

plugins/modules/k8s_service.py
```diff
@@ -195,7 +196,7 @@ def merge_dicts(x, y):
             if isinstance(x[k], dict) and isinstance(y[k], dict):
                 yield (k, dict(merge_dicts(x[k], y[k])))
             else:
-                yield (k, y[k])
+                yield (k, x[k] if x[k] else y[k])
```
gravesm, line 199: This would be a breaking change, wouldn't it? I'm also not sure what bug it would be addressing.

abikouo: This is not a breaking change, the module is not working as expected. Here we are trying to merge the definition built using modules parameters and the one from the ``src`` parameter.
Supposed the service port is not set on module parameter, then we are going to merge

(definition from ``src`` parameter)
```
(...)
spec:
   port: xxx
```

definition from module parameters
```
(...)
spec:
   port: None
```
The result will be 
```
(...)
spec:
   port: None
```
The API will complain when trying to create the service, as the port is mandatory.
This fix simply adds a precedence on the value found from the ``src`` option

gravesm: Could you add a bugfix section to the changelog to document this? The change here would need to be `y[k] if y[k] else x[k]`, though. The module params currently take precedence over the resource definition.

gravesm (approved):
Thanks! LGTM

plugins/action/k8s_info.py
```diff
@@ -359,7 +359,12 @@ def run(self, tmp=None, task_vars=None):
         # find the file in the expected search path
         src = self._task.args.get("src", None)
 
-        if src:
+        if (
+            src
+            and not src.startswith("http://")
```
Akasurde, line 364:
```suggestion
            and not src.startswith(("http://", "https://", "ftp://")):
```
`startswith` takes a tuple as well. Not a blocker, just a hint for the future.

Akasurde (approved):
Rest LGTM

maxpain:
When we should wait for the release with this feature?

PR #489, opened 2022-07-17
feat: add new tail_lines parameter to k8s_log module (#488)

### LINKED ISSUE #488: k8s_log: add parameter to limit the number of log lines to be retrieved
##### SUMMARY

In the current implementation for `k8s_log`, there is no parameter to limit the number of log lines to be retreived.

That parameter is helpful to reduce resource consumption and for easier log handling, as `--tail=n` option for `kubectl`.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

`kubernetes.core.k8s_log`

##### ADDITIONAL INFORMATION

As a workaround for the current implementation, the number of logs can be limited by specifying number like `{{ _logs.log_lines[-10:] }}` when handling logs. However, it is more reasonable to retrieve only the specified number of logs in the first place, as in the following example.

```yaml
- name: Retrieve logs
  kubernetes.core.k8s_log:
    api_version: apps/v1
    kind: Deployment
    namespace: default
    name: sample-app
    tail_lines: 10     👈👈👈
  register: _deployment_logs
```

kurokobo:
##### SUMMARY

Add new `tail_lines` parameter to `k8s_log` module to limit the number of lines to be retrieved from the end of the logs.

Closes #488.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

`kubernetes.core.k8s_log`

##### ADDITIONAL INFORMATION

Thanks for useful collection 😃 
This is the first time to send PR to this collection, so please let me know if I'm on the wrong way.

- The `version_added` is set to `2.4.0`, but I'm not aware of the roadmap for this collection, so I'd like to know this is the right version to specify.
- Changelog and simple integration test is also added.
- It seems that the end of `log_lines` always contains an empty element, so if `tail_lines` is set to `5`, the length of `log_lines` will be 6, as noted in the comment in the test. I've considered that truncating the trailing empty element, but decided not to for the following reasons.
  - It is inconsistent and unnatural to remove trailing empty elements only when tail_lines is specified.
  - Removing trailing empty elements always with or without tail_lines is a destructive change and should not be done because it would break backward compatibility.

Example tasks in playbook:

```yaml
  tasks: 
    - name: create a job that has 10 lines of log
      kubernetes.core.k8s:
        state: present
        wait: yes
        wait_timeout: "{{ k8s_wait_timeout | default(omit) }}"
        wait_condition:
          type: Complete
          status: 'True'
        definition:
          apiVersion: batch/v1
          kind: Job
          metadata:
            name: multiline-log
            namespace: test
          spec:
            template:
              spec:
                containers:
                  - name: busybox
                    image: busybox
                    command: ['sh']
                    args: ['-c', 'for i in $(seq 0 9); do echo $i; done']
                restartPolicy: Never
            backoffLimit: 4

    - name: retrieve all logs from the job
      kubernetes.core.k8s_log:
        api_version: batch/v1
        kind: Job
        namespace: test
        name: multiline-log
      register: full_log

    - name: retrieve last 5 lines of log from the job
      kubernetes.core.k8s_log:
        api_version: batch/v1
        kind: Job
        namespace: test
        name: multiline-log
        tail_lines: 5
      register: tailed_log

    - ansible.builtin.debug:
        var: full_log.log_lines

    - ansible.builtin.debug:
        var: tailed_log.log_lines
```

Example output:

```bash
TASK [create a job that has 10 lines of log] *****************************************************************************************
ok: [localhost]

TASK [retrieve all logs from the job] ************************************************************************************************
ok: [localhost]

TASK [retrieve last 5 lines of log from the job] *************************************************************************************
ok: [localhost]

TASK [ansible.builtin.debug] *********************************************************************************************************
ok: [localhost] => 
  full_log.log_lines:
  - '0'
  - '1'
  - '2'
  - '3'
  - '4'
  - '5'
  - '6'
  - '7'
  - '8'
  - '9'
  - ''

TASK [ansible.builtin.debug] *********************************************************************************************************
ok: [localhost] => 
  tailed_log.log_lines:
  - '5'
  - '6'
  - '7'
  - '8'
  - '9'
  - ''

PLAY RECAP ***************************************************************************************************************************
localhost                  : ok=5    changed=0    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0 
```

### DIFF
--- a/changelogs/fragments/488-add-support-for-tail-logs.yaml
+++ b/changelogs/fragments/488-add-support-for-tail-logs.yaml
@@ -1,0 +1,3 @@
+---
+minor_changes:
+- k8s_log - added the `tail_lines` parameter to limit the number of lines to be retrieved from the end of the logs (https://github.com/ansible-collections/kubernetes.core/issues/488).
--- a/plugins/modules/k8s_log.py
+++ b/plugins/modules/k8s_log.py
@@ -67,6 +67,12 @@ options:
     type: bool
     default: False
     version_added: '2.4.0'
+  tail_lines:
+    description:
+    - A number of lines from the end of the logs to retrieve.
+    required: no
+    type: int
+    version_added: '2.4.0'
 
 requirements:
   - "python >= 3.6"
@@ -106,6 +112,7 @@ EXAMPLES = r"""
     kind: DeploymentConfig
     namespace: testing
     name: example
+    tail_lines: 100
   register: log
 """
 
@@ -156,6 +163,7 @@ def argspec():
             since_seconds=dict(),
             label_selectors=dict(type="list", elements="str", default=[]),
             previous=dict(type="bool", default=False),
+            tail_lines=dict(type="int"),
         )
     )
     return args
@@ -203,6 +211,11 @@ def execute_module(svc, params):
 
     if params.get("previous"):
         kwargs.setdefault("query_params", {}).update({"previous": params["previous"]})
+
+    if params.get("tail_lines"):
+        kwargs.setdefault("query_params", {}).update(
+            {"tailLines": params["tail_lines"]}
+        )
 
     response = resource.log.get(
         name=name, namespace=namespace, serialize=False, **kwargs
--- a/tests/integration/targets/k8s_log/tasks/main.yml
+++ b/tests/integration/targets/k8s_log/tasks/main.yml
@@ -112,6 +112,46 @@
       assert:
         that: job_logs.log_lines[0] == "7"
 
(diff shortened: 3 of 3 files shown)


kurokobo:
The cause of the two FAILUREs both appear to be `kubernetes.core.kubectl` connection plugin and appear to totally be unrelated to the changes made in my PR :thinking:

abikouo (approved):

gravesm (approved):
Thanks for the PR. The failures are for known issues in development versions of ansible, so nothing to worry about.

PR #495, opened 2022-07-26
Update sanity ignores

gravesm:
##### SUMMARY

A recent change in CI config means we need to add ignores for
unsupported versions of python (2.6, 2.7 and 3.5). These changes
effectively prevent sanity from doing anything when running on those
versions and Ansible < 2.12. In Ansible 2.12 and later, this can be
handled on a global level in tests/config.yml.

##### ISSUE TYPE

- Bugfix Pull Request
- Docs Pull Request
- Feature Pull Request
- New Module Pull Request

##### ADDITIONAL INFORMATION

```paste below

```


abikouo (approved):

gravesm:
regate

gravesm:
regate

PR #497, opened 2022-08-01
helm - add support for in-memory kubeconfig

### LINKED ISSUE #492: Add non file based kubeconfig option to helm modules in kubernetes.core
First Time posting an issue here, I believe I tracked down where this was discussed before, and this may just be an oversight, or a difference in how the helm plugins work from the rest of the k8s plugins. Please correct me if I have missed something important.

##### SUMMARY
Apply ability to use non file based kubeconfig to remaining kubernetes core modules to 

Per Issue #139 Which was already resolved and merged for the kubernetes.core.k8s module, The ability to use a non file based kubeconfig should be generalized to the other core modules. 

See the already merged pull request #212 

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
These components do not have the fix applied, and would benefit from the change.
kubernetes.core.helm 
kubernetes.core.helm_info
kubernetes.core.helm_plugin
kubernetes.core.helm_plugin_info
kubernetes.core.kubectl

also, although it works for kubernetes.core.k8s, the documentation on the 
kubernetes.core.k8s inventory and lookup documentation links under Inventory Plugins and Lookup Plugins appear to be out of date, not stating that it can be used as a dict with the constraints

##### ADDITIONAL INFORMATION
The previous ticket has justifications, but the simplest example is the ability to store kubeconfig objects in an external secret store. Leveraging something like the hashi_vault module to deposit them inline, without having to distribute secrets onto users machines directly to use them is a much cleaner solution.

since this features works for many of the kubernetes.core modules, it would be very beneficial to have a consistent expectation of the kubeconfig field accross all the modules in kubernetes.core.

abikouo:
##### SUMMARY

closes #492 

##### ISSUE TYPE

- Feature Pull Request

### DIFF
--- a/changelogs/fragments/497-helm-add-support-for-in-memory-kubeconfig.yml
+++ b/changelogs/fragments/497-helm-add-support-for-in-memory-kubeconfig.yml
@@ -1,2 +1,2 @@
 minor_changes:
-  - helm, helm_plugin, helm_info, helm_plugin_info - add support for in-memory kubeconfig. (https://github.com/ansible-collections/kubernetes.core/issues/492).
+  - helm, helm_plugin, helm_info, helm_plugin_info, kubectl - add support for in-memory kubeconfig. (https://github.com/ansible-collections/kubernetes.core/issues/492).
--- a/changelogs/fragments/497-helm-add-support-for-in-memory-kubeconfig.yml
+++ b/changelogs/fragments/497-helm-add-support-for-in-memory-kubeconfig.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - helm, helm_plugin, helm_info, helm_plugin_info - add support for in-memory kubeconfig. (https://github.com/ansible-collections/kubernetes.core/issues/492).
--- a/plugins/module_utils/helm_args_common.py
+++ b/plugins/module_utils/helm_args_common.py
@@ -1,0 +1,44 @@
+from __future__ import absolute_import, division, print_function
+
+from ansible.module_utils.basic import env_fallback
+
+__metaclass__ = type
+
+
+HELM_AUTH_ARG_SPEC = dict(
+    binary_path=dict(type="path"),
+    context=dict(
+        type="str",
+        aliases=["kube_context"],
+        fallback=(env_fallback, ["K8S_AUTH_CONTEXT"]),
+    ),
+    kubeconfig=dict(
+        type="raw",
+        aliases=["kubeconfig_path"],
+        fallback=(env_fallback, ["K8S_AUTH_KUBECONFIG"]),
+    ),
+    host=dict(type="str", fallback=(env_fallback, ["K8S_AUTH_HOST"])),
+    ca_cert=dict(
+        type="path",
+        aliases=["ssl_ca_cert"],
+        fallback=(env_fallback, ["K8S_AUTH_SSL_CA_CERT"]),
+    ),
+    validate_certs=dict(
+        type="bool",
+        default=True,
+        aliases=["verify_ssl"],
+        fallback=(env_fallback, ["K8S_AUTH_VERIFY_SSL"]),
+    ),
+    api_key=dict(
+        type="str",
+        no_log=True,
+        fallback=(env_fallback, ["K8S_AUTH_API_KEY"]),
+    ),
+)
+
+HELM_AUTH_MUTUALLY_EXCLUSIVE = [
+    ("context", "ca_cert"),
+    ("context", "validate_certs"),
+    ("kubeconfig", "ca_cert"),
+    ("kubeconfig", "validate_certs"),
+]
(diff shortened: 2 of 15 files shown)


abikouo:
recheck

abikouo:
recheck

plugins/module_utils/helm_args_common.py
```diff
@@ -0,0 +1,44 @@
...
+        fallback=(env_fallback, ["K8S_AUTH_API_KEY"]),
+    ),
+)
+
+HELM_AUTH_MUTUALLY_EXCLUSIVE = [
+    ("context", "ca_cert"),
+    ("context", "validate_certs"),
+    ("kubeconfig", "ca_cert"),
+    ("kubeconfig", "validate_certs"),
+]
```
gravesm, line 44: This is a breaking change since `helm` does not currently set `kubeconfig` and `validate_certs` as mutually exclusive. To be honest, I'm not sure that any of these should be mutually exclusive, but if we want to change that, we should deal with it in a separate PR and it needs to go through the normal deprecation process.

plugins/modules/helm_info.py
```diff
@@ -176,63 +182,34 @@ def get_release_status(module, command, release_name, release_state):
     return release
 
 
+def argument_spec():
+    arg_spec = copy.deepcopy(HELM_AUTH_ARG_SPEC)
+    arg_spec.update(
+        dict(
+            release_name=dict(type="str", required=True, aliases=["name"]),
+            release_namespace=dict(type="str", required=True, aliases=["namespace"]),
+            release_state=dict(type="list", default=[], elements="str"),
```
gravesm, line 191: `release_state` is not a parameter for this module.

plugins/modules/helm_plugin.py
```diff
@@ -132,60 +135,32 @@ def main():
...
+    module = AnsibleModule(
+        argument_spec=argument_spec(),
         supports_check_mode=True,
         required_if=[
             ("state", "present", ("plugin_path",)),
             ("state", "absent", ("plugin_name",)),
             ("state", "latest", ("plugin_name",)),
         ],
-        mutually_exclusive=[
-            ("plugin_name", "plugin_path"),
```
gravesm, line 170: You have removed this from the module.

gravesm (commented):
In addition to these changes, this PR does not fully address the linked issue, as it does not add this functionality to the inventory plugin, the connection plugin or the lookup plugin. I think we can skip the inventory plugin as that will be deprecated in the next version. I'm not sure about the value of supporting this in the lookup plugin, as passing a dictionary in a lookup plugin seems pretty unwieldy, but I guess if someone wants to implement it we can consider it. It seems reasonable to support this for the connection plugin, though. If you don't want to do that in this PR, could you please either leave that issue open and add a note to it, or create a new issue for the pieces that weren't implemented?

plugins/modules/helm_info.py
abikouo: `release_state` is well defined as parameter of this module, 
https://github.com/ansible-collections/kubernetes.core/blob/5ff3566f308f07aabd7a0347a438eab17cb6fedb/plugins/modules/helm_info.py#L42-L55

plugins/module_utils/helm_args_common.py
abikouo: I will keep it as it is for now, if needed could be changed later

abikouo:
> In addition to these changes, this PR does not fully address the linked issue, as it does not add this functionality to the inventory plugin, the connection plugin or the lookup plugin. I think we can skip the inventory plugin as that will be deprecated in the next version. I'm not sure about the value of supporting this in the lookup plugin, as passing a dictionary in a lookup plugin seems pretty unwieldy, but I guess if someone wants to implement it we can consider it. It seems reasonable to support this for the connection plugin, though. If you don't want to do that in this PR, could you please either leave that issue open and add a note to it, or create a new issue for the pieces that weren't implemented?

I will add support for the connection plugin, it sounds reasonnable

plugins/modules/helm_info.py
gravesm: Sorry, I have no idea what I was looking at. You are correct.

changelogs/fragments/497-helm-add-support-for-in-memory-kubeconfig.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - helm, helm_plugin, helm_info, helm_plugin_info - add support for in-memory kubeconfig. (https://github.com/ansible-collections/kubernetes.core/issues/492).
```
gravesm, line 2:
```suggestion
  - helm, helm_plugin, helm_info, helm_plugin_info, kubectl - add support for in-memory kubeconfig. (https://github.com/ansible-collections/kubernetes.core/issues/492).
```

gravesm (commented):
Thanks, one small change, but otherwise looks good.

gravesm (approved):

PR #500, opened 2022-08-10
Honor aliases for lookup and inventory plugins

abikouo:
rebase and extend the following PR #71 

##### ISSUE TYPE

- Bugfix Pull Request


gravesm (approved):
Thanks!

PR #505, opened 2022-08-23
Add example usage of from_yaml_all

badnetmask:
Depends-On: https://github.com/ansible-collections/kubernetes.core/pull/513

##### SUMMARY
Sometimes one might want to use a single YAML file that contains multiple Kubernetes definitions. This PR updates the documentation to provide a simple example of how to accomplish that.

##### ISSUE TYPE
- Docs Pull Request

##### COMPONENT NAME
k8s

##### ADDITIONAL INFORMATION
I have not tested this solution against other modules, simply because I did not have use-case for those.

### DIFF
--- a/changelogs/fragments/505-add-from-yaml-all-example.yml
+++ b/changelogs/fragments/505-add-from-yaml-all-example.yml
@@ -1,0 +1,3 @@
+---
+trivial:
+  - add example usage for from_yaml_all (https://github.com/ansible-collections/kubernetes.core/pull/505).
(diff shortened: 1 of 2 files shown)


gravesm:
@badnetmask There's no need to use a loop to do this. The `definition` parameter will already accept a list of resources. You can just do:
```yaml
- kubernetes.core.k8s:
    state: present
    definition: "{{ lookup('file', 'myresources.yml') | from_yaml_all }}"
```

badnetmask:
> @badnetmask There's no need to use a loop to do this. The `definition` parameter will already accept a list of resources. You can just do:

Well, I'm glad it works without a loop (and I have updated the PR to reflect that), but I still find useful to call out the example, since multiple definitions in a single file appears to be a very common pattern.

gravesm:
recheck

goneri:
recheck

goneri:
recheck

gravesm (approved):

gravesm:
regate

Issue #491, opened 2022-07-21
Add force option to helm_repository


gygitlab:
##### SUMMARY
Sometimes a Helm repo needs to be updated with a new URL. The `helm repo add` command allows for this with the `--force-update` flag:

```
      --force-update               replace (overwrite) the repo if it already exists
```

Ask is for this option to be available on the Ansible task.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

`kubernetes.core.helm_repository`

##### ADDITIONAL INFORMATION

To update an existing repo.

```yaml
- name: Add a repository
  kubernetes.core.helm_repository:
    name: stable
    repo_url: https://kubernetes.github.io/ingress-nginx
    force: true
```

mor120:
Can I take it If decided to add this option?

gravesm:
@mor120 If you would like to implement this, that would be great.

mor120:
Starting to work on it then :)

mor120:
Branch:
https://github.com/mor120/kubernetes.core/tree/helm-repo-force-update

mor120:
PR: https://github.com/ansible-collections/kubernetes.core/pull/509

PR #509, opened 2022-09-09
[helm] Add the force_update arg

mor120:
Depends-On: https://github.com/ansible/ansible-zuul-jobs/pull/1648
Depends-On: https://github.com/ansible-collections/kubernetes.core/pull/522

##### SUMMARY
Sometimes a Helm repo needs to be updated with a new URL. The `helm repo add` command allows for this with the `--force-update` flag:
```
      --force-update               replace (overwrite) the repo if it already exists
```

##### ISSUE TYPE
- Feature Pull Request - Closes https://github.com/ansible-collections/kubernetes.core/issues/491

##### COMPONENT NAME
`kubernetes.core.helm_repository`

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/plugins/modules/helm_repository.py
+++ b/plugins/modules/helm_repository.py
@@ -112,6 +112,13 @@ options:
     type: raw
     aliases: [ kubeconfig_path ]
     version_added: "2.4.0"
+  force_update:
+    description:
+    - Whether or not to replace (overwrite) the repo if it already exists.
+    type: bool
+    aliases: [ force ]
+    default: False
+    version_added: "2.3.3"
 """
 
 EXAMPLES = r"""
@@ -218,6 +225,7 @@ def install_repository(
     repository_username,
     repository_password,
     pass_credentials,
+    force_update
 ):
     install_command = command + " repo add " + repository_name + " " + repository_url
 
@@ -227,6 +235,9 @@ def install_repository(
 
     if pass_credentials:
         install_command += " --pass-credentials"
+
+    if force_update:
+      install_command += " --force-update"
 
     return install_command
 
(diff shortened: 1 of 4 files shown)


mor120:
Going to write test and validate.

plugins/modules/helm_repository.py
```diff
@@ -97,6 +97,13 @@
     type: path
     aliases: [ ssl_ca_cert ]
     version_added: "2.3.0"
+  force_update:
+    description:
+    - Whether or not to replace (overwrite) the repo if it already exists.
+    type: bool
+    aliases: [ force ]
+    default: False
+    version_added: "2.3.3"
```
mor120, line 106: What version should I set here?

plugins/modules/helm_repository.py
```diff
@@ -275,14 +292,15 @@ def main():
         helm_cmd = delete_repository(helm_cmd, repo_name)
         changed = True
     elif repo_state == "present":
-        if repository_status is None:
+        if repository_status is None or force_update:
```
mor120, line 295: My concern is that even if the specs are exactly the same, we will still mark it as changed, on the other side, in helm it works the same, for example:
```
➜ helm repo add stable https://kubernetes.github.io/ingress-nginx
"stable" already exists with the same configuration, skipping
➜ helm repo add stable https://kubernetes.github.io/ingress-nginx --force-update
"stable" has been added to your repositories
```

tests/integration/targets/helm/tasks/main.yml
```diff
@@ -4,4 +4,4 @@
   loop_control:
     loop_var: helm_version
   with_items:
-    - "v3.2.4"
+    - "v3.7.0"
```
mor120, line 7: To get the tests passed.

plugins/modules/helm_repository.py
gravesm: You should use `2.4.0` here.

mor120: Done, thanks.

plugins/modules/helm_repository.py
mor120: Alright, thanks.

gravesm (approved):

gravesm:
regate

goneri:
recheck

PR #510, opened 2022-09-09
k8s_exec - add support for stdin

### LINKED ISSUE #506: Support stdin in k8s_exec
##### SUMMARY
Currently there is no way to to pass anything via stdin using the k8s_exec module, since Stdin is explicitly disabled [1]
Both command and shell ansible builtins support a stdin: parameter which could be used to pass data to a command inside container.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
k8s_exec

##### ADDITIONAL INFORMATION
Currently if you want to inject data inside a container with ansible, most of the times you need to resort to using approaches like:
`echo foo | kubectl exec -it -n ns pod -- sh -c "cat - > /tmp/foo-content"`

k8s_cp is very often a no-go as it requires tar to be installed inside the container, which has become less and less likely as people try to slim down containers all the time.

```yaml
kubernetes.core.k8s_exec:
  namespace: "{{ vault_ns }}"
  pod: "{{ vault_pod }}"
  command:  bash -e -c "cat - > /tmp/destfile"
  stdin: "{{ data_to_inject }}"
```

[1] https://github.com/ansible-collections/kubernetes.core/blob/main/plugins/modules/k8s_exec.py#L191-L202

abikouo:
closes #506 

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

k8s_exec


gravesm (commented):
I spent some time testing this and I'm not sure if we can really support this feature. I guess kubernetes does not provide any way to close stdin on an exec call with websockets. Using the following task:
```yaml
    - kubernetes.core.k8s_exec:
        namespace: default
        pod: mypod
        command: /bin/sh -e -c "cat - | does_not_exist"
        stdin: "some content"
```
sometimes this will correctly fail with an error message that the command doesn't exist, and sometimes it will claim to have succeeded. Closing the connection entirely is the only way to close stdin but I'm guessing this means that sometimes the response is truncated since we're forcing it. The only way I can get it to consistently fail is to put a `time.sleep(1)` at the top of the `while resp.is_open()` loop. I don't really think this is a workable solution. It's possible that maybe there's a bug in the python kubernetes implementation, but I spent some time looking through it and couldn't find one.

abikouo:
> I spent some time testing this and I'm not sure if we can really support this feature. I guess kubernetes does not provide any way to close stdin on an exec call with websockets. Using the following task:
> 
> ```yaml
>     - kubernetes.core.k8s_exec:
>         namespace: default
>         pod: mypod
>         command: /bin/sh -e -c "cat - | does_not_exist"
>         stdin: "some content"
> ```
> 
> sometimes this will correctly fail with an error message that the command doesn't exist, and sometimes it will claim to have succeeded. Closing the connection entirely is the only way to close stdin but I'm guessing this means that sometimes the response is truncated since we're forcing it. The only way I can get it to consistently fail is to put a `time.sleep(1)` at the top of the `while resp.is_open()` loop. I don't really think this is a workable solution. It's possible that maybe there's a bug in the python kubernetes implementation, but I spent some time looking through it and couldn't find one.

This is something I have identified too during the implementation, it makes sense to me to not support this feature
We need to clarify before how to handle errors in a deterministic way

abikouo:
closing for now, waiting update on `kubernetes-client`

PR #518, opened 2022-09-23
k8s_rollback/tests: speed up test

goneri:
Set a low timeout for the tasks that are expected to fail fast to speed up the whole tests.


changelogs/fragments/k8s_rollback_reduce_tmeouts.yaml
```diff
@@ -0,0 +1,3 @@
+---
+minor_changes:
```
gravesm, line 2:
```suggestion
trivial:
```

gravesm (approved):

PR #531, opened 2022-10-18
Added possibility to get all values by helm_info module

Fors1:
##### SUMMARY
Parameter **get_all_values** has been added, which is passed to function get_values. Default is **False**. Parameter is **not** required.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
helm_info

##### ADDITIONAL INFORMATION
Unfortunately, helm_info module lacks functionality of getting all the values of a helm release, including the default ones. This restricts upgrade and config migration capabilities. Parameter **get_all_values** has been added. This parameter, if set, adds `-a` parameter to `helm get values` call. The parameter is not required and defaults to False, so backwards compability is complied.


plugins/modules/helm_info.py
```diff
@@ -53,6 +53,12 @@
     type: list
     elements: str
     version_added: "2.3.0"
+  get_all_values:
+    description:
+      - Set to true if you want to get all (computed) values of the release.
+    required: false
+    default: false
+    type: bool
```
gravesm, line 61:
```suggestion
    type: bool
    version_added: "2.4.0"
```

plugins/modules/helm_info.py
```diff
@@ -145,7 +151,7 @@ def get_release(state, release_name):
 
 
 # Get Release state from deployed release
-def get_release_status(module, command, release_name, release_state):
+def get_release_status(module, command, release_name, release_state, get_all_values=False):
```
gravesm, line 154:
```suggestion
def get_release_status(
    module, command, release_name, release_state, get_all_values=False
):
```

gravesm (commented):
Thanks for the PR! A few small changes. This will also need [a changelog fragment](https://docs.ansible.com/ansible/latest/community/development_process.html#community-changelogs).

plugins/modules/helm_info.py
```diff
@@ -53,6 +53,14 @@
     type: list
     elements: str
     version_added: "2.3.0"
+  get_all_values:
+    description:
+      - Set to C(True) if you want to get all (computed) values of the release.
```
abikouo, line 58:
```suggestion
      - Set to I(True) if you want to get all (computed) values of the release.
```

plugins/modules/helm_info.py
```diff
@@ -53,6 +53,14 @@
...
+      - When C(False) (default), only user supplied values are returned.
```
abikouo, line 59:
```suggestion
      - When I(False) (default), only user supplied values are returned.
```

abikouo (approved):
@Fors1 thanks for contributing to this collection!!
Few minor changes, otherwise code looks good to me

plugins/modules/helm_info.py
gravesm: `I()` is for option names and `C()` is for option values, so what is here is correct.

https://docs.ansible.com/ansible/latest/dev_guide/developing_modules_documenting.html#linking-and-other-format-macros-within-module-documentation

plugins/modules/helm_info.py
gravesm: Same as above. `C()` is correct.

gravesm (approved):

PR #561, opened 2023-01-03
fix multiple issues with dry_run logic

abikouo:
##### SUMMARY

Fix multiple issues with dry_run logic
 - The parameter value passed to the client set to `dry_run=All` instead of `dry_run=True`.
 - Add conditional check for Kubernetes release for the dry_run to be set
 - Add integration test that checks to ensure server side dry run is being used during check mode.

##### ISSUE TYPE

- Bugfix Pull Request


gravesm (approved):
Thanks, this looks good!

jillr (approved):

PR #575, opened 2023-01-30
helm - add reuse_values and reset_values support

abikouo:
##### SUMMARY

closes #394 

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

helm
##### ADDITIONAL INFORMATION

```paste below

```

PR #587, opened 2023-02-18
Add validate argument to helm_template

pauvos:
##### SUMMARY

Support `helm template --validate`:

```
--validate

validate your manifests against the Kubernetes cluster you are currently pointing at. This is the same validation performed on an install
```

It's sometimes necessary to render templates against the kubernetes API because helm charts increasingly start to use conditions:

```mustache
{{- if and (.Capabilities.APIVersions.Has "networking.k8s.io/v1") (semverCompare ">= v1.19.0-0" .Capabilities.KubeVersion.Version) -}}
  {{- $apiVersion = "networking.k8s.io/v1" -}}
{{- else if .Capabilities.APIVersions.Has "networking.k8s.io/v1beta1" -}}
  {{- $apiVersion = "networking.k8s.io/v1beta1" -}}
{{- else -}}
  {{- $apiVersion = "extensions/v1beta1" -}}
{{- end }}
apiVersion: {{ $apiVersion }}
kind: Ingress
```

Without checking the capabilities of the cluster the above template will render a deprecated extensions/v1beta1 Ingress resource.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

kubernetes.core.helm_template

##### ADDITIONAL INFORMATION

```paste below

```


pauvos:
I would like to add a unittest but I couldn't figure out how to run them.

Does this make sense?

```python
# tests/unit/modules/test_helm_template.py

def test_template_with_validate():
    my_chart_ref = "testref"
    helm_cmd = "helm"
    parser = argparse.ArgumentParser()
    # to "simulate" helm template options, include two optional parameters NAME and CHART.
    # if parsed string contains only one parameter, the value will be passed
    # to CHART and NAME will be set to default value "release-name" as in helm template
    parser.add_argument("cmd")
    parser.add_argument("template")
    parser.add_argument("NAME", nargs="?", default="release-name")
    parser.add_argument("CHART", nargs="+")
    parser.add_argument("--validate", dest="validate", action="store_true")
    parser.set_defaults(validate=False)

    mytemplate = template(cmd=helm_cmd, chart_ref=my_chart_ref, validate=True)

    args, unknown = parser.parse_known_args(mytemplate.split())

    assert args.validate is True
```

plugins/modules/helm_template.py
```diff
@@ -99,6 +99,12 @@
     type: list
     elements: str
     version_added: 2.4.0
+  validate:
+    description:
+      - Validate your manifests against the Kubernetes cluster you are currently pointing at. This is the same validation performed on an install.
+    required: false
+    type: bool
+    default: false
```
abikouo, line 107:
```suggestion
    default: false
    version_added: 2.5.0
```

abikouo (changes requested):
@pauvos thanks for your pull request, 
could you please add a changelog? here is an example https://github.com/ansible-collections/kubernetes.core/pull/575/files#diff-6d789f5197c05bc24880f186dfc451b420fc61f8666d969d2c7309473aa1e3f1

pauvos:
@abikouo , yes sure thing!

I have 2 other pull requests:
* #586
* #588 

Should i merge everything in one pull requests or create individual changelog fragments?

plugins/modules/helm_template.py
pauvos: forced pushed a changelog fragment and added `version_added`

gravesm (commented):
Thanks for the PR, and sorry for taking so long to get back to this. Since this is going to connect to a cluster, you'll need to add some connection parameters to the module. This should be pretty easy, though, and you can look at how the helm module does it. I think it should be sufficient to just extend the doc string as it's done in https://github.com/ansible-collections/kubernetes.core/blob/8640c16cd41c3713d33fdcb97ba022b7587c4988/plugins/modules/helm.py#L221-L222 and then add those args to the arg spec for the module, similar to how it's done in https://github.com/ansible-collections/kubernetes.core/blob/deb4859f193c12c535e1451992d8913753fa131e/plugins/modules/helm.py#L695.

Since this is adding functionality to connect to a cluster, it would be a good idea to add an integration test or two for this. You can add those in https://github.com/ansible-collections/kubernetes.core/blob/8640c16cd41c3713d33fdcb97ba022b7587c4988/tests/integration/targets/helm/tasks/tests_chart.yml#L345 and just use that as a guide.

Your suggestion of the unit test seems reasonable. If you want to run unit tests locally you can do so with `ansible-test units --docker`.

willthames:
There is an alternative way to achieve the desired results here which is to add an api versions parameter to the helm_template module (this is often useful when there isn't a kubernetes cluster to connect to e.g. during CI - this is how we generate all of our templates)

PR #608, opened 2023-04-04
Fix broken idempotence when using Helm '--post-renderer' module arg

ekwisnek:
SUMMARY
The post_renderer module arg is not being used for helmdiff, only for helm_deploy. This change fixes that.

ISSUE TYPE
Bugfix Pull Request

COMPONENT NAME
kubernetes.core.helm

ADDITIONAL INFORMATION
The "bug" here is that, when using the '--post-renderer' module arg, idempotence is broken because the post_renderer arg isn't being used in the diff.


ekwisnek:
@gravesm Thanks, Mike!

PR #643, opened 2023-09-04
Extend hidden_fields to allow more complicated field definitions

willthames:
##### SUMMARY
This allows us to ignore e.g. the last-applied-configuration annotation by specifying
`metadata.annotations[kubectl.kubernetes.io/last-applied-configuration]`

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
hidden_fields

### DIFF
--- a/plugins/module_utils/k8s/service.py
+++ b/plugins/module_utils/k8s/service.py
@@ -501,23 +501,21 @@ def diff_objects(
     result["before"] = diff[0]
     result["after"] = diff[1]
 
-    if list(result["after"].keys()) != ["metadata"] or list(
+    if list(result["after"].keys()) == ["metadata"] and list(
         result["before"].keys()
-    ) != ["metadata"]:
-        return False, result
+    ) == ["metadata"]:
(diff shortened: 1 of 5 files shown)


willthames:
Note that I haven't added a changelog fragment as we haven't done a release since the PR introducing this feature and the previous fragment is still valid for this change:

```
minor_changes:
  - k8s, k8s_info - add a hidden_fields option to allow fields to be hidden in the results of k8s and k8s_info
```

willthames:
Note that it's currently impossible to meet both black and pep8 with regard to list slices

black wants `list[left : right]`

pep8 wants `list[left:right]`

More things seem to complain about pep8.

willthames:
I don't really understand why all the integration tests are failing, it doesn't seem that it's related to this PR (but happy to fix if I'm wrong about that)

willthames:
The failing sanity tests on milestone and devel also seem unrelated to this PR (suspect a breaking change in devel ansible which probably needs separate investigation - has the yamllint sanity check changed?)

PR #655, opened 2023-11-06
Cleanup GitHub workflows

GomathiselviS:
##### SUMMARY

This PR pins ansible-lint version to 6.21.x and organizes the workflows based on https://github.com/ansible-collections/cloud-content-handbook/blob/main/CI/README.md.

Refer https://issues.redhat.com/browse/ACA-1117

##### ISSUE TYPE

- Bugfix Pull Request
- Docs Pull Request
- Feature Pull Request
- New Module Pull Request

##### ADDITIONAL INFORMATION

```paste below

```


.github/workflows/sanity-test.yaml
```diff
@@ -0,0 +1,92 @@
...
+      - stable-*
+    tags:
+      - '*'
+
+jobs:
+  sanity:
+    uses: ansible-network/github_actions/.github/workflows/sanity.yml@main
+    with:
+      matrix_include: "[]"
+      matrix_exclude: >-
```
gravesm, line 23: I feel like we should be able to update the ansible-network workflow to only test 2.14+ at this point, which should remove the need for us to maintain any kind of excludes here.

.github/workflows/linters.yaml
```diff
@@ -0,0 +1,28 @@
...
+    tags:
+      - '*'
+
+jobs:
+  linters:
+    uses: ansible-network/github_actions/.github/workflows/tox-linters.yml@main
+  ansible-lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: ansible-network/github_actions/.github/actions/checkout_dependency@main
```
hakbailey, line 25: Just curious why we are using this instead of the regular `checkout` action?

GomathiselviS:
recheck

.ansible-lint
```diff
@@ -0,0 +1,6 @@
+---
+profile: production
+
+exclude_paths:
+  - tests/integration
+  - tests/sanity
```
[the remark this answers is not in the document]
GomathiselviS: ansible-lint scans the ignore-*.txt files and throws 'sanity[cannot-ignore]' error

.github/workflows/changelog.yaml
```diff
@@ -0,0 +1,23 @@
...
+    types:
+      - opened
+      - reopened
+      - labeled
+      - unlabeled
+      - synchronize
+    branches:
+      - main
+      - stable-*
+    tags:
```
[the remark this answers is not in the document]
GomathiselviS: As per my understanding, with tags set to * ,  this action will be triggered when a new tag is created.

.github/workflows/galaxy-import.yaml
```diff
@@ -0,0 +1,20 @@
+name: galaxy-import
+concurrency:
+  group: ${{ github.head_ref }}
+  cancel-in-progress: true
+
+on:
+  pull_request:
+    types:
```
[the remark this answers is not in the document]
GomathiselviS: I shall remove them.

plugins/modules/helm_info.py
```diff
@@ -137,6 +137,7 @@
       version_added: "2.4.0"
 """
 
+# testing
```
alinabuzachis, line 140: Is this comment needed?

plugins/inventory/k8s.py
```diff
@@ -98,27 +98,27 @@
...
+  connections:
+    - host: https://192.168.64.4:8443
+      api_key: xxxxxxxxxxxxxxxx
+      validate_certs: false
+
+- name: Use default config (~/.kube/config) file and active context, and return objects for a specific namespace
+  plugin: kubernetes.core.k8s
+  connections:
+    - namespaces:
     - testing
```
alinabuzachis, line 115: This should be intended probably?!

.github/workflows/changelog.yaml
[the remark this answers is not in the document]
hakbailey: I don't think it works that way under the `pull_request` event. I only see in the docs that `tag` can be used with the `push` event which makes sense because you don't do a PR for a tag, you just push it. Regardless, I'm not sure we want to run all (or maybe any?) of these workflows when a tag is added...changelog in particular doesn't seem applicable to tagging.

.github/workflows/integration-tests.yaml
```diff
@@ -0,0 +1,137 @@
...
+      test_targets: ${{ needs.splitter.outputs.test_targets }}
+    runs-on: ubuntu-latest
+    timeout-minutes: 60
+    strategy:
+      fail-fast: false
+      matrix:
+        ansible-version:
+          - milestone
+        python-version:
+          - "3.11"
```
hakbailey, line 53: This should probably be 3.12 since that's the most recent supported version.

plugins/inventory/k8s.py
GomathiselviS: Ansible-lint version 6.21 examines the indentation in the EXAMPLES block. While it could be skipped, I chose not to do so because the issue was observed in only a few files.

.github/workflows/linters.yaml
```diff
@@ -0,0 +1,28 @@
...
+      - synchronize
+    branches:
+      - main
+      - stable-*
+    tags:
+      - '*'
+
+jobs:
+  linters:
+    uses: ansible-network/github_actions/.github/workflows/tox-linters.yml@main
```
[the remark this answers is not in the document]
GomathiselviS: @hakbailey I've incorporated the isort command. Addressing the errors thrown by mypy will require more effort, and I propose tackling it post-release. Additionally, since we need to implement it across all our repositories, what are your thoughts on taking the task after the release?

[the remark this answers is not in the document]
hakbailey: I think that's probably fine. Just out of curiosity, does it still give a lot of errors if you do something similar to what's already implemented in the [cloud.terraform collection](https://github.com/ansible-collections/cloud.terraform/blob/8c52bf2a0a0f5e3ba56f57d0ac5bf1dd8efd0788/tox.ini#L32)?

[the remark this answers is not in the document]
gravesm: Yes, there are definitely ways we can configure mypy to limit the amount of upfront work that's needed, but we should limit this PR to fixing only what's necessary right now. Our goal is just to get the CI to a state that we can actually merge things. We can focus on improving it later.

.github/workflows/sanity-tests.yaml
```diff
@@ -0,0 +1,15 @@
...
+
+on:
+  pull_request:
+    branches:
+      - main
+      - stable-*
+
+jobs:
+  sanity:
+    uses: GomathiselviS/github_actions/.github/workflows/sanity.yml@remove_2_9
```
gravesm, line 15: This needs to be changed to use ansible-network.

.github/workflows/unit-tests.yaml
```diff
@@ -0,0 +1,14 @@
...
+
+on:
+  pull_request:
+    branches:
+      - main
+      - stable-*
+
+jobs:
+  unit-source:
+    uses: GomathiselviS/github_actions/.github/workflows/unit_source.yml@remove_2_9
```
gravesm, line 14: Same here, change to use ansible-network.

changelogs/fragments/gha-organize-workflows.yaml
```diff
@@ -0,0 +1,3 @@
+---
+trivial:
+  - organize the workflows and pin ansible-lint to 6.21.x.
```
gravesm, line 3: I think since we're changing the meta/runtime.yml to require ansible 2.14+ we should mark this PR as a breaking change.

gravesm (approved):

gravesm (approved):

plugins/inventory/k8s.py
alinabuzachis: I mean - testing should be indented. It should be a namespace, or am I wrong.

jillr (approved):

abikouo (approved):

tox.ini
```diff
@@ -27,19 +30,33 @@ commands =
...
-  yamllint
+  {[testenv:black]deps}
+  {[testenv:isort]deps}
   flake8
-  {[testenv:black_check]deps}
-
+  yamllint
 commands =
-  {[testenv:black_check]commands}
+  black -v --check {toxinidir}/plugins {toxinidir}/tests
```
abikouo, line 59:
```suggestion
  flynt {[common]format_dirs}
  black -v --check {toxinidir}/plugins {toxinidir}/tests
```

tox.ini
```diff
@@ -27,19 +30,33 @@ commands =
...
+[testenv:flynt]
+deps =
+  flynt
+commands =
+  flynt {[common]format_dirs}
 
 [testenv:linters]
 deps =
-  yamllint
+  {[testenv:black]deps}
```
abikouo, line 54:
```suggestion
  {[testenv:black]flynt}
  {[testenv:black]deps}
```

abikouo (commented):
Should we not use the `flyint` for the linters ?

plugins/inventory/k8s.py
gravesm: Yeah, @alinabuzachis is correct. I just pushed a fix.

tox.ini
gravesm: If we want to use flynt we should add it to the linters in https://github.com/ansible-collections/cloud-content-handbook/tree/main/CI. For this PR, we should merge this as is. We can look at adding flynt later.

hakbailey (approved):

alinabuzachis (approved):

abikouo:
@gravesm removing the `mergeit` and adding `test-all-the-targets` label to see if the full integration tests suite is passing.
if ok, I will add it back.

.github/workflows/integration-tests.yaml
```diff
@@ -0,0 +1,135 @@
+name: Integration tests
+on:
+  pull_request:
```
abikouo, line 3:
```suggestion
  pull_request:
     types:
       - opened
       - reopened
       - labeled
       - unlabeled
       - synchronize
```

PR #725, opened 2024-05-22
Prepare release 4.0.0

abikouo:
##### SUMMARY

- Update galaxy.yml version
- Run antsibull to generate release content

##### ISSUE TYPE

- Release Pull Request


gravesm (changes requested):
This needs to be filed against the stable-4 branch, not main. Also, if there are linting changes needed, they should be first added to main in a separate PR, then backported.

PR #731, opened 2024-05-29
kubevirt.core collection cross testing

abikouo:
##### SUMMARY

The `kubevirt.core` collection has dependency with this collection. We define new workflows to ensure that nothing is broken on that collection when pushing new changes on this collection.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

``CI``

### DIFF
--- a/tools/kubevirt_list_targets.py
+++ b/tools/kubevirt_list_targets.py
@@ -1,0 +1,22 @@
+import os
+import sys
+from pathlib import PosixPath
+
+
+def main():
+
+    src = sys.argv[1]
+    path = PosixPath(src) / PosixPath("tests/integration/targets/")
+
+    def _is_disable(path):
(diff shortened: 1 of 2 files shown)


gravesm (approved):

gravesm:
@abikouo Is it possible to make this non-blocking for now?

tools/kubevirt_list_targets.py
```diff
@@ -0,0 +1,22 @@
+import os
```
alinabuzachis, line 1: Can this file be moved into the gitHub_actions repo? What do you think?

gravesm: I think we should keep this here for now. There's nothing else that's going to be using it.

alinabuzachis (approved):

PR #745, opened 2024-06-10
helm: Accept release candidate versions for compatibility checks

gileri:
##### SUMMARY

If the helm CLI version includes `-rc.1` for example, the version checks fails due to an incomplete regex.
The error can be triggered if you use [helm v3.15.0-rc.1](https://github.com/helm/helm/releases/tag/v3.15.0-rc.1) for example, and [apply a helm chart with `wait: true` ](https://github.com/ansible-collections/kubernetes.core/blob/main/plugins/modules/helm.py#L806-L807)

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME
helm
helm_pull

##### ADDITIONAL INFORMATION

### DIFF
--- a/changelogs/fragments/20240611-helm-rc-version.yaml
+++ b/changelogs/fragments/20240611-helm-rc-version.yaml
@@ -1,0 +1,2 @@
+bugfixes:
+  - helm - Helm version checks did not support RC versions. They now accept any version tags. (https://github.com/ansible-collections/kubernetes.core/pull/745).
--- a/plugins/module_utils/helm.py
+++ b/plugins/module_utils/helm.py
@@ -184,10 +184,10 @@ class AnsibleHelmModule(object):
     def get_helm_version(self):
         command = self.get_helm_binary() + " version"
         rc, out, err = self.run_command(command)
-        m = re.match(r'version.BuildInfo{Version:"v([0-9\.]*)",', out)
+        m = re.match(r'version.BuildInfo{Version:"v([^"]*)",', out)
         if m:
             return m.group(1)
-        m = re.match(r'Client: &version.Version{SemVer:"v([0-9\.]*)", ', out)
+        m = re.match(r'Client: &version.Version{SemVer:"v([^"]*)", ', out)
         if m:
             return m.group(1)
         return None
--- a/plugins/module_utils/helm.py
+++ b/plugins/module_utils/helm.py
@@ -184,10 +184,10 @@ class AnsibleHelmModule(object):
     def get_helm_version(self):
         command = self.get_helm_binary() + " version"
         rc, out, err = self.run_command(command)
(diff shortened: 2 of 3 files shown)


gravesm:
recheck

gravesm:
recheck

changelogs/fragments/20240611-helm-rc-version.yaml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - helm - Helm version checks did not support RC versions. They now accept any version tags. (https://github.com/ansible-collections/kubernetes.core/pull/745).
```
yurnov, line 2: Please add [new_line_at_end_of_file](https://yamllint.readthedocs.io/en/stable/rules.html#module-yamllint.rules.new_line_at_end_of_file)

gileri: I think there is one, GH perhaps don't show it :

```
% cat -e changelogs/fragments/20240611-helm-rc-version.yaml
bugfixes:$
  - helm - Helm version checks did not support RC versions. They now accept any version tags. (https://github.com/ansible-collections/kubernetes.core/pull/745).$
```

```
% yamllint changelogs/fragments/20240611-helm-rc-version.yaml && echo 'ok'
ok
```

Lint passed too. Am I missing something ?

yurnov: Probably I missed it in review, but I did see it [here](https://github.com/gileri/kubernetes.core/blob/helm-version-rc/changelogs/fragments/20240117-fix-helm-diff-cmd-line-rendering.yml). Anyway, it's ok if linter passed

gileri: Github does display an explicit symbol when there is **no** newline at end-of-file.

yurnov: > Github does display an explicit symbol when there is **no** newline at end-of-file.

got it, thanks

yurnov (approved):

plugins/module_utils/helm.py
```diff
@@ -184,10 +184,10 @@ def get_helm_binary(self):
     def get_helm_version(self):
         command = self.get_helm_binary() + " version"
         rc, out, err = self.run_command(command)
-        m = re.match(r'version.BuildInfo{Version:"v([0-9\.]*)",', out)
+        m = re.match(r'version.BuildInfo{Version:"v([^"]*)",', out)
```
gravesm, line 187: I'd suggest getting rid of the character set and using a non-greedy operator:
```suggestion
        m = re.match(r'version.BuildInfo{Version:"v(.*?)",', out)
```

gileri: Done.

gravesm (approved):

PR #795, opened 2024-11-19
k8s_cluster_info.py Add warn for resorce missing kind field

AnthonyWC:
##### SUMMARY
 Add warn for resorce missing kind field

##### ISSUE TYPE
Currently k8s_cluster_info.py just assume kind field will always exists; this helps to warn/debug issue when resource is missing kind field

### DIFF
--- a/plugins/modules/k8s_cluster_info.py
+++ b/plugins/modules/k8s_cluster_info.py
@@ -178,6 +178,12 @@ def execute_module(module, client):
             if resource.group == ""
             else "/".join([resource.group, resource.group_version.split("/")[-1]])
         )
+
+        kind = getattr(resource, "kind", None)
+        if not kind:
+            module.warn(f"Resource is missing 'kind': {resource}")
+            continue
+
         results[key][resource.kind] = {
             "categories": resource.categories if resource.categories else [],
             "name": resource.name,


plugins/modules/k8s_cluster_info.py
gravesm, line 1: Have you found a particular case where there's an error? As far as I'm aware this shouldn't be necessary. The check should already be happening in https://github.com/kubernetes-client/python/blob/07fb11f2bc6a6356854e9dabcd7d65a62e849feb/kubernetes/base/dynamic/resource.py#L29

AnthonyWC: yes; specifically i ran into this while using kiali-operator: https://github.com/kiali/kiali-operator/blob/master/roles/default/kiali-deploy/tasks/main.yml#L23

But i imagine it's generally a good idea to check/warn for exception as it is used in many places.  In general, cluster api should almost always have kind field but for whatever reason I have one k8 cluster where it failed.

gravesm: Can you reproduce the error? I'm not opposed to adding a check if it's necessary, but it doesn't look to me like your change would fix that problem.

AnthonyWC: My change isn't meant to fix the error but to warn on which resource caused the error.  The fix should be applied on kubernetes api side.

PR #873, opened 2025-02-03
Fix linters in CI

yurnov:
##### SUMMARY
It seems that recent updates in linters break CI. Closes #874

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
CI

PR #875, opened 2025-02-05
updating helm version regex for more support

BlankCanvasStudio:
##### SUMMARY
Not all versions of helm output their version info with the format:

```
version.BuildInfo{Version:"vX.X.X", ....}
```

some (specifically 3.16.4) use:

```
version.BuildInfo{Version:"X.X.X", ....}
```

This causes this function to return `None`, instead of the version string. I've update the regex to support either format.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME

This is an update to `plugins/module_utils/helm.py`

##### ADDITIONAL INFORMATION

When using the output of `get_helm_version` to create a `LooseVersion`:

https://github.com/ansible-collections/kubernetes.core/blob/main/plugins/module_utils/_version.py#L306

If the version does not match the regex, `None` will be returned, but the object will still be created. 

If the object is then compared with a pinned version, like in this example:

https://github.com/ansible-collections/kubernetes.core/blob/main/plugins/modules/helm.py#L808

without verifying the "version" attribute exists (ie verifying `get_helm_version` returned a string), the program will crash.

`LooseVersion` could also be updated to throw an error when None is passed to its constructor and/or `get_helm_version` could be updated to throw an error when the program can't regex for the system version.

### DIFF
--- a/plugins/module_utils/helm.py
+++ b/plugins/module_utils/helm.py
@@ -185,7 +185,7 @@ class AnsibleHelmModule(object):
     def get_helm_version(self):
         command = self.get_helm_binary() + " version"
         rc, out, err = self.run_command(command)
-        m = re.match(r'version.BuildInfo{Version:"v(.*?)",', out)
+        m = re.match(r'version.BuildInfo{Version:"v?(.*?)",', out)
         if m:
             return m.group(1)
         m = re.match(r'Client: &version.Version{SemVer:"v(.*?)", ', out)


gravesm (commented):
Can you add a unit test for this in https://github.com/ansible-collections/kubernetes.core/blob/main/tests/unit/module_utils/test_helm.py?

gravesm:
Ah, also this will need a [changelog fragment](https://docs.ansible.com/ansible/latest/community/development_process.html#community-changelogs). Ignore the failing linters for now. #873 should fix those.

BlankCanvasStudio:
@gravesm I just added a unit test. It seems from the unit tests that returning None from get_helm_version() is intentional behavior, but there aren't any type checks on its result in the code base. Is this the intended behavior? This issue will keep popping up unless either get_helm_version explicitly creates an error or if type checks are added. I personally think it should generate an error, but I'm wondering why that wasn't the design choice when this was built. I feel like I'm missing something.

yurnov:
> @gravesm I'm also confused about adding a changelog fragment. The docs you link say I should add a file to `changelogs/fragments/`, but I don't see that directory in the repository. Are they included from somewhere else? Further, should I also do a version bump in `changelog.yaml`?

It seems that the version issue is the same as #875

Regarding changelog fragments, just create this folder and place a fragment for your change, you can use 9bdb9a297c5a2612871c1ac7e601179dcad75d88 as the example.

PR #884, opened 2025-03-14
Added create_only param to ignore existing objects

dcraigmv:
##### SUMMARY
Instead of checking for existing resources before creating a new resource to avoid modifying it each time e.g.

```
k8s_info:
  kind: Secret
  name: my-auto-generated-psk
  namespace: project-1
register: existing_secret

k8s:
  state: present
  definition:
    kind: Secret
    metadata:
      name: my-auto-generated-psk
      namespace: project-1
    data:
      admin: "{{ lookup('community.general.random_string', length=15, base64=true) }}"
when: existing_secret.resources == []
register: new_secret

set_fact:
  # check for existing and new resource, consolidate into one fact for later use
  admin_psk: "{{ (existing_secret.resources | default([new_secret.result])) | first }}"
```

This new `create_only` param reduces this to a single task

```
k8s:
  state: present
  create_only: true
  definition:
    kind: Secret
    metadata:
      name: my-auto-generated-psk
      namespace: project-1
    data:
      admin: "{{ lookup('community.general.random_string', length=15, base64=true) }}"
register: admin_psk
```

Otherwise this task would recreate the secret's data each time. This is especially useful with `register` as the original resource can be used in subsequent tasks without checking for existing or new. It is also beneficial in ansible operators.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
k8s create_only param

##### ADDITIONAL INFORMATION
A couple of unit test cases added

### DIFF
--- a/plugins/module_utils/k8s/runner.py
+++ b/plugins/module_utils/k8s/runner.py
@@ -175,6 +175,10 @@ def perform_action(svc, definition: Dict, params: Dict) -> Dict:
         if params.get("apply"):
             instance, warnings = svc.apply(resource, definition, existing)
             result["method"] = "apply"
+        elif existing and params.get("create_only", False):
+            instance = existing.to_dict()
+            result["changed"] = False
+            result["method"] = "create"
         elif not existing:
             if state == "patched":
                 result.setdefault("warnings", []).append(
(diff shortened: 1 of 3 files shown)


dcraigmv:
`immutable` causes a failure, but this can be used with `immutable` to avoid that.

PR #888, opened 2025-03-20
Run integration tests using ansible-core 2.19

abikouo:
Validate Integration tests targets against fallible ansible version

##### ISSUE TYPE

- Feature Pull Request


mandar242 (approved):

alinabuzachis (approved):

plugins/action/k8s_info.py
gravesm, line 1: Why all these changes? Can't you just use `trust_value()` as outlined in the [porting guide](https://github.com/nitzmahone/ansible-documentation/blob/core_data_tagging/docs/docsite/rst/porting_guides/porting_guide_core_2.19.rst#applying-template-trust-to-individual-values).

abikouo: @gravesm The first fix was to apply `trust_as_template()` to template resource, however, this is not working for the omit use case. Consider the following use case where we have the following definition.

```
apiVersion: v1
metadata:
   namespace: "{{ test_namespace | default(omit) }}"
```

When applying trust as a template to the following data, Ansible will return. 

```
apiVersion: v1
metadata:
   namespace: ""
```
The templating module basically returns `""` as a value for Omit, this is because the string is passed as a raw buffer. A possible solution will be pushing `dict` value to the templating mechanisms, however this wont be possible because the `trust_as_template` does not accept `dict` variables. The only solution there is then to go through the full document and push single elements to the templating module

beeankha (approved):

PR #915, opened 2025-04-28
fix(k8s,service): Hide fields first before creating diffs

0xFelix:
##### SUMMARY

By hiding fields first before creating diff hidden fields will not be shown in the results diffs and therefore will also not trigger the changed condition.

The issue can only be reproduced when a mutating webhook changes the object while the kubernetes.core.k8s module is working with it.

https://github.com/kubevirt/kubevirt.core/issues/145

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

kubernetes.core.module_utils.k8s.service

##### ADDITIONAL INFORMATION

Run kubernetes.core.k8s and create object with hidden fields. After run kubernetes.core.k8s again and let a webhook mutate the object that the module is working with. The module should return with `changed: no`.

```paste below

```


0xFelix:
@abikouo @gravesm Can you help with this?

abikouo (approved):

changelogs/fragments/20250428-k8s-service-hide-fields-first.yaml
```diff
@@ -0,0 +1,3 @@
+---
+minor_changes:
```
gravesm, line 2:
```suggestion
bugfixes:
```

0xFelix:
Updated to address your comments

0xFelix:
Good question, is this wanted behavior or a bug?

https://github.com/ansible-collections/kubernetes.core/blob/914a16ec5cfb806ccc529d610295c9481da4566b/tests/integration/targets/k8s_hide_fields/tasks/main.yml#L61

0xFelix:
@gravesm Can you shed some light on this? Is this an acceptable change (still a bugfix?)?

0xFelix:
@abikouo @gravesm @yurnov Can you help to move this forward? Is this an acceptable change? Would it still be considered a bugfix and be backportable?

abikouo:
> Good question, is this wanted behavior or a bug?
> 
> https://github.com/ansible-collections/kubernetes.core/blob/914a16ec5cfb806ccc529d610295c9481da4566b/tests/integration/targets/k8s_hide_fields/tasks/main.yml#L61

@0xFelix this has to be fixed. Even if the field is hidden, a change to the resource should still result in a change.

0xFelix:
@abikouo That's what this PR is about, changing the behavior of `hide_fields` so hidden fields do not trigger a change anymore. Why would I be interested in a change to a hidden field if I can't see it anyway?

0xFelix:
Why do `metadata.resourceVersion` and `metadata.generation` don't trigger a changed condition then? I don't think it is desirable to always have the changed condition. Isn't that what `hide_fields` should be for, allowing me to ignore changes to certain fields because I'm not interested in them?

0xFelix:
Also, the use case in this PR is a bit tricky. The change condition is not triggered by a field I'm setting explicitly, but rather because a mutating webhook changed something during the attempted but ultimately not required update to a resource (it was already in the desired state).

A concrete example is `kubemacpool` on OCP, which adds/updates a timestamp when it last processed an object to the objects annotations. As a user I never set this and I'm not interested in this annotation at all, yet I'm never able to run a task without getting the changed condition because of the value changing each time.

0xFelix:
@abikouo Not sure how the error in `helm_registry_auth` is related to this, but I did not touch it and the code looks not to be importing the changed code either.

yurnov:
Hi @0xFelix

I checked this issue and found a root cause; it's not related to your PR and will be fixed separately. Details added here https://github.com/ansible-collections/kubernetes.core/issues/944#issuecomment-2940891033

yurnov:
Integration test failed due to rate-limit, @0xFelix, could you please push empty commit in 20-60 min to trigger tests once again?

0xFelix:
@abikouo Not sure what these new failures are about, broken helm tests?

yurnov:
This time failed part:
```
TASK [helm : Assert that helm is not installed] ********************************
Tuesday 08 July 2025  15:47:18 +0000 (0:00:00.489)       0:00:14.623 ********** 
[ERROR]: Task failed: Action failed: Assertion failed
Origin: /home/runner/collections/ansible_collections/kubernetes/core/tests/integration/targets/helm/tasks/test_helm_not_installed.yml:11:3

 9   register: helm_missing_binary
10
11 - name: Assert that helm is not installed
     ^ column 3

    "assertion": "'No such file or directory' in helm_missing_binary.msg",
    "changed": false,
    "evaluated_to": false,
    "msg": "Assertion failed"
}
```

It's a code:
```
- name: Assert that helm is not installed
  assert:
    that:
      - helm_missing_binary is failed
      - "'No such file or directory' in helm_missing_binary.msg"
```

0xFelix:
@yurnov @beeankha Thank you for taking care of it!

PR #919, opened 2025-04-28
Update ansible-lint version to 25.1.2

beeankha:
Update ansible-lint version to 25.1.2


yurnov:
I think that changelog can be skipped here (just add label `skip-changelog`) as it a CI change that not included to the collection in Galaxy.

changelogs/fragments/919-update-ansible-lint-version.yaml
```diff
@@ -0,0 +1,3 @@
+---
```
yurnov, line 1: it's not requried here as it a CI-only change

abikouo (approved):

PR #933, opened 2025-05-16
[DNM] Prep release `6.0.0`

beeankha:
##### SUMMARY

Prep `kubernetes.core 6.0.0`

**Prerequisite:** Release of [`community.okd`](https://github.com/openshift/community.okd)/`redhat.openshift` `4.0.2` needs to happen first

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME
Multiple

PR #941, opened 2025-06-03
Update `dev` version to `7.0.0-dev0` after `6.0.0` release

beeankha:
##### SUMMARY
Updating the `dev` version listed in `Makefile` and `galaxy.yml` since [6.0.0](https://github.com/ansible-collections/kubernetes.core/pull/933) has been released.


yurnov:
I think that you can just update versions as an additional commit to #942, but two PR in sequence will be fine either

PR #989, opened 2025-08-13
Changelog fixes

alinabuzachis:
##### SUMMARY

Changelog fixes

##### ISSUE TYPE

- Docs Pull Request

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -60,20 +60,20 @@ v5.3.0
 Release Summary
 ---------------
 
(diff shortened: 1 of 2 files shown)


beeankha:
It looks like `stable-5` only needs the change from `changelog.yaml`, the `CHANGELOG.rst` file is formatted properly on that branch:

Screenshot from 2025-08-14 09-57-47

I can make a quick small PR for `stable-5` so that everything matches up and we don't have to worry about manual backporting/conflicts since it's just one file that needs updating.

PR #1016, opened 2025-10-07
Update `main` after `6.2.0` release

beeankha:


README.md
```diff
@@ -99,7 +99,7 @@ You can also include it in a `requirements.yml` file and install it via `ansible
 ---
 collections:
   - name: kubernetes.core
-    version: 6.1.0
+    version: 6.2.0
```
abikouo, line 102: I'm not sure how we should handle this value. The galaxy.yml version has the value `7.0.0-dev0` while the README.md will have `6.2.0`

beeankha: I think the standard is for us to list the latest released version in the README for `main` (vs the `*-dev` version), so this value should be OK

GomathiselviS: AFAIU, in the dev branch (in this case, main), the version specified in galaxy.yml indicates that the collection is under development for the upcoming 7.0.0 release , not the final published version. The version in the README.md, on the other hand, reflects what users currently see or use. Please correct me if I’m mistaken.

yurnov: Hi @abikouo, I think is should be fine, in the Galaxy we have `7.0.0-dev0` to easy see that it's not released `latest` version from the master in the output of `ansible-galaxy collection list`, and README.md is more for the users that looking to GitHub for some documentation/advice/whatever, and in this place it better to have "latest released" version. More experianced users will find how to install the collection from the master without guide in readme

Issue #1046, opened 2025-12-17
Return value key ... should not be used for return values since it cannot be accessed with dot notation in Jinja


hdefazio:
##### SUMMARY
The `validate-modules:bad-return-value-key` check is failing in the devel sanity checks in the CI 
Ref: https://github.com/ansible/ansible/pull/86079 
This was also reported + fixed in amazon.aws: https://github.com/ansible-collections/amazon.aws/pull/2772 

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
validate-modules

##### ANSIBLE VERSION

```
devel

```

##### COLLECTION VERSION

```
latest (main branch) 

```

##### OS / ENVIRONMENT
Github CI

##### STEPS TO REPRODUCE
Run the sanity tests in the CI

##### EXPECTED RESULTS
All CI is passing

##### ACTUAL RESULTS
```
ERROR: Found 3 validate-modules issue(s) which need to be resolved:
ERROR: plugins/modules/helm.py:0:0: bad-return-value-key: Return value key 'values' found in status should not be used for return values since it cannot be accessed with dot notation in Jinja
ERROR: plugins/modules/helm_info.py:0:0: bad-return-value-key: Return value key 'values' found in status should not be used for return values since it cannot be accessed with dot notation in Jinja
ERROR: plugins/modules/k8s.py:0:0: bad-return-value-key: Return value key 'items' found in result should not be used for return values since it cannot be accessed with dot notation in Jinja
See documentation for help: https://docs.ansible.com/ansible-core/devel/dev_guide/testing/sanity/validate-modules.html

```

PR #1051, opened 2025-12-21
add validate-modules:bad-return-value-key to ignore list

yurnov:
##### SUMMARY
Return value key 'values', 'items', and 'keys' are not recommended for module return keys, and starting from 2.21, they will raise a warning in the ansible-test, ref. https://github.com/ansible/ansible/pull/86079

The proper way is to rename return values in the modules:
- plugins/modules/helm.py
- plugins/modules/helm_info.py
- plugins/modules/k8s.py

However, to keep backward compatibility, we need to keep them in the current `stable-6` and `stable-5` branches, and this commit is to stabilize CI for the time being.

A CI-only commit does not require a changelog. Tags `skip-changelog`, `backport-6`, and `backport-5` to be added.
  
Resolve: #1050

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
CI only

Issue #1050, opened 2025-12-21
[CI sanity] ansible-test validate-modules sanity test: report ‘bad’ return values in `devel`


yurnov:
##### SUMMARY

Current CI failed on `devel` with any PR with messages like:
```
(bad-return-value-key: Return value key 'values' found in status should not be used for return values since it cannot be accessed with dot notation in Jinja)
```

[Example of fail](https://github.com/ansible-collections/kubernetes.core/actions/runs/20398813964/job/58618185153#step:10:243)

This is related to https://github.com/ansible/ansible/pull/86079

A proper fix requires renaming return keys `values` and `items`, that is, **backward incompatible** changes, so a deprecation period and a deprecation warning must be initiated, and the actual change requires a new major release. As the solution for the current `stable-5` and `stable-6` a adding `validate-modules:bad-return-value-key` to `tests/sanity/ignore-2.21.txt` is proposed.

Reference: [Ansible forum](https://forum.ansible.com/t/ansible-test-validate-modules-sanity-test-report-bad-return-values-that-cannot-be-accessed-with-jinjas-dot-notation/44833)

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
CI (sanity test)
plugins/modules/helm.py
plugins/modules/helm_info.py
plugins/modules/k8s.py

##### ANSIBLE VERSION
`devel` (2.21)

##### COLLECTION VERSION

main
stable-6
stable-5

PR #1056, opened 2026-01-08
address sanity issues

abikouo:
##### SUMMARY

- helm/helm_info - Deprecate some parameters and add new ones to resolve sanity issues.
- k8s - the return block doc is not aligned with what the module returns

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

helm, helm_info, k8s

Fixes: #1046

### DIFF
--- a/plugins/modules/helm.py
+++ b/plugins/modules/helm.py
@@ -419,14 +419,14 @@ status:
       returned: always
       description: The Date of last update
     values:
-      type: str
+      type: dict
       returned: always
       description:
         - Dict of Values used to deploy.
         - This return value has been deprecated and will be removed in a release after
           2027-01-08. Use RV(status.release_values) instead.
     release_values:
-      type: str
+      type: dict
       returned: always
       description:
         - Dict of Values used to deploy.
--- a/plugins/modules/helm_info.py
+++ b/plugins/modules/helm_info.py
@@ -115,14 +115,14 @@ status:
       returned: always
       description: The Date of last update
     values:
-      type: str
+      type: dict
       returned: always
       description:
         - Dict of Values used to deploy
         - This return value has been deprecated and will be removed in a release after
           2027-01-08. Use RV(status.release_values) instead.
     release_values:
-      type: str
+      type: dict
       returned: always
       description:
         - Dict of Values used to deploy.


changelogs/fragments/20260108-fix-sanity-failures.yml
```diff
@@ -0,0 +1,11 @@
+---
+bugfixes:
+  - k8s - Fix return block from the module documentation ().
+deprecated_features:
+  - helm - the ``status.values`` return value has been deprecated and will be removed in a release after 207-01-08.
```
yurnov: 207?

yurnov:
Hi @abikouo,

I mentioned in the comment to #1051 that your change is the proper solution to #1050, and I fully support that, but an important point is that it's backward-incompatible changes that users do not expect in the minor release (6.3.0 and 5.5.0), so, probably better is to add `deprecated` to the `RETURN` on the module itself to have upcoming backward incompatible changes documented in 6.3.0 (and 5.5.0, i.e. add in into #1051) and introduce them in the version 7.0.0?

What do you think?

att @abikouo @beeankha @felixfontein

felixfontein:
When deprecating a return value, you can simply mention the deprecation in the return value's description and the changelog. This is easily overlooked though, and since ansible-core 2.19 you can use Data Tagging to mark a return value as deprecated. If someone uses that value in a role/playbook, they will see a deprecation message there!

I wrote a module util that allows to use this also with older ansible-core versions (there it's simply a no-op): https://github.com/ansible-collections/community.hrobot/blob/main/plugins/module_utils/_tagging.py

plugins/modules/helm.py
```diff
@@ -421,7 +421,16 @@
     values:
       type: str
       returned: always
-      description: Dict of Values used to deploy
+      description:
+        - Dict of Values used to deploy.
+        - This return value has been deprecated and will be removed in a release after
+          2027-01-08. Use RV(status.release_values) instead.
+    release_values:
+      type: str
```
beeankha: Should this be `type: dict`?

plugins/modules/helm.py
```diff
@@ -421,7 +421,16 @@
     values:
       type: str
```
beeankha: Same here, should this be `type: dict`?

plugins/modules/helm_info.py
```diff
@@ -117,7 +117,16 @@
     values:
       type: str
```
beeankha: This might need to be `type: dict`

plugins/modules/helm_info.py
```diff
@@ -117,7 +117,16 @@
...
       returned: always
-      description: Dict of Values used to deploy
+      description:
+        - Dict of Values used to deploy
+        - This return value has been deprecated and will be removed in a release after
+          2027-01-08. Use RV(status.release_values) instead.
+    release_values:
+      type: str
```
beeankha: Ditto here, `type: dict`.

PR #1086, opened 2026-02-03
Add `check_mode` support for `k8s_drain` module

### LINKED ISSUE #1037: drain module should not warn, or have a nowarn, when user requested the action
##### SUMMARY
Drain module should have an option to disable warnings about things that were explicitly requested by the user.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
k8s_drain

##### ADDITIONAL INFORMATION
These warnings are tautological, because the user has already had to provide a non-default option to request the action, so it warns about something that is expected.  I would argue they should default to off anyways or be removed, but it's debatable.  At least an option would be good.  For example this makes little sense:

```
kubernetes.core.k8s_drain:
  ...
  delete_options:
    ignore_daemonsets: true
    delete_emptydir_data: true
```
gets the result:
```
[WARNING]: Deleting Pods with local storage: kube-system/cilium-cst6l.
[WARNING]: Ignoring DaemonSet-managed Pods: kube-system/cilium-envoy-wcpd6.
```

I have specifically requested these things, yet it warns me.  Ok maybe for verbose safety purpose, but not for ordinary operation, at least have a `warn` boolean.  Although kubectl warns about this, I find that silly too, at least allow to suppress these in the ansible module so the tooling output is not polluted.

Note: action warnings can be suppressed in ansible.cfg or in environment, but only on a global basis, so this is best implemented as module arguments.

abikouo:
##### SUMMARY

Closes #1037
- added support for `check_mode`
- Converted warnings into informational display when user has explicitly requested to delete daemontset-managed pods, unmanaged pods or pods with local storage

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

`k8s_drain`

### DIFF
--- a/changelogs/fragments/20260203-k8s_drain-warning-fixes.yaml
+++ b/changelogs/fragments/20260203-k8s_drain-warning-fixes.yaml
@@ -1,5 +1,5 @@
 ---
 minor_changes:
-  - k8s_drain - added support for check_mode (https://github.com/ansible-collections/kubernetes.core/pull/1086).
-  - k8s_drain - Converted module warnings into informational displays when users explicitly request the deletion of
-    unmanaged pods, pods with local storage, or those managed by a DaemonSet. (https://github.com/ansible-collections/kubernetes.core/issues/1037).
+  - k8s_drain - Add support for `check_mode` (https://github.com/ansible-collections/kubernetes.core/pull/1086).
+  - k8s_drain - Convert module warnings into informational displays when users explicitly request the deletion of
+    unmanaged pods, pods with local storage, or those managed by a `DaemonSet` (https://github.com/ansible-collections/kubernetes.core/issues/1037).


changelogs/fragments/20260203-k8s_drain-warning-fixes.yaml
```diff
@@ -0,0 +1,5 @@
+---
+minor_changes:
+  - k8s_drain - added support for check_mode (https://github.com/ansible-collections/kubernetes.core/pull/1086).
+  - k8s_drain - Converted module warnings into informational displays when users explicitly request the deletion of
+    unmanaged pods, pods with local storage, or those managed by a DaemonSet. (https://github.com/ansible-collections/kubernetes.core/issues/1037).
```
beeankha:
```suggestion
  - k8s_drain - Add support for `check_mode` (https://github.com/ansible-collections/kubernetes.core/pull/1086).
  - k8s_drain - Convert module warnings into informational displays when users explicitly request the deletion of
    unmanaged pods, pods with local storage, or those managed by a `DaemonSet` (https://github.com/ansible-collections/kubernetes.core/issues/1037).
```

tests/integration/targets/k8s_drain/tasks/drain.yml
```diff
@@ -0,0 +1,389 @@
...
+    that:
+    - drain_force_check_mode is changed
+    - '"node "+node_to_drain+" marked unschedulable." in drain_force_check_mode.result'
+
+- name: Ensure node remains schedulable
+  ansible.builtin.include_tasks: tasks/validate_node_status.yml
+  vars:
+    schedulable: true
+
+- name: Assert that running with check_mode did not deleted any Pod
```
beeankha:
```suggestion
- name: Assert that running with check_mode did not delete any Pod
```

beeankha (approved):
Other than some minor wording/typo changes, this looks good to me!

PR #1108, opened 2026-04-21
ci: remove cloud.common and turbo mode from testing matrix

yurnov:
##### SUMMARY
The cloud.common collection is incompatible with ansible-core >= 2.19.0. With the current testing matrix using Python 3.12 and the ansible milestone (currently 2.22), this incompatibility causes integration tests to fail.

This commit removes the installation of the cloud.common collection and removes turbo mode from the testing matrix to fix CI test runs.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME

`.github/workflows/integration-tests.yaml`

##### ADDITIONAL INFORMATION

Fixes the issue that led to failed CI in #1101

PR #1101, opened 2026-04-17
Release prep for 6.4.0

beeankha:
##### SUMMARY

Prep kubernetes.core `6.4.0` release

##### COMPONENT NAME

Multiple

PR #1109, opened 2026-04-21
ci: conditionally test turbo mode and cloud.common

yurnov:
##### SUMMARY
The cloud.common collection is incompatible with ansible-core >= 2.19.0. With the current testing matrix using Python 3.12 and the ansible milestone (currently 2.22), this incompatibility causes integration tests to fail.

Instead of [completely removing turbo mode from the testing matrix](https://github.com/ansible-collections/kubernetes.core/pull/1108), this PR adds ansible-core 2.18 to the matrix and excludes the combination of the ansible milestone and turbo mode. The checkout and installation of the cloud.common collection are now conditionally executed only when turbo mode is enabled.

It keeps testing of turbo mode and also testing of milestone for the ensure compatiability with the upcoming version of ansible-core. 

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME

`.github/workflows/integration-tests.yaml`

##### ADDITIONAL INFORMATION

Fixes the issue that led to failed CI in #1101. Alternative approach of #1108

### DIFF
--- a/.github/workflows/integration-tests.yaml
+++ b/.github/workflows/integration-tests.yaml
@@ -55,13 +55,13 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ansible-version:
-          - milestone
+        ansible-version: ["2.18", "milestone"]
+        enable-turbo-mode: [true, false]
+        exclude:
+          - ansible-version: "milestone"
+            enable-turbo-mode: true
         python-version:
           - "3.12"
-        enable-turbo-mode:
-          - true
-          - false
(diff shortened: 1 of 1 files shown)


yurnov:
Hi @beeankha @GomathiselviS,

Just to be on the same page, PRs #1108 and 1109 are two alternative approaches to the same; when one is accepted, the other will be abandoned. I think that this one should be accepted and #1108 abandoned, but keep both approaches to allow the maintainers to decide.

abikouo (approved):

beeankha (approved):
Thank you for doing this work, @yurnov !

PR #1140, opened 2026-06-05
Support wait_for_jobs flag for helm release upgrade/install

dansimko:
##### SUMMARY

The `helm` module missing the option to specify `--wait-for-jobs` flag in addition to the already supported `--wait`.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME
helm

##### ADDITIONAL INFORMATION

The helm cli supports the `--wait-for-jobs` flag that waits for all jobs to complete before marking the release as successful. We have a use-case where we perform database migrations and database seeding using a job during release deployment. We need to wait for the job to finish successfully or roll back the deployment (i.e. we need `--atomic` along with `--wait-for-jobs`). This flag was not supported by the `helm` module.

### DIFF
--- a/plugins/modules/helm.py
+++ b/plugins/modules/helm.py
@@ -187,6 +187,12 @@ options:
         It will wait for as long as I(wait_timeout). This feature requires helm>=3.7.0. Added in version 2.3.0.
     default: False
     type: bool
+  wait_for_jobs:
+    description:
+      - When I(release_state) is set to C(present), and I(wait) is set to C(True), wait until all jobs are in a
+        completed state before marking the release as successful. Added in version 6.5.0.
+    default: False
+    type: bool
   wait_timeout:
     description:
       - Timeout when wait option is enabled (helm2 is a number of seconds, helm3 is a duration).
@@ -385,6 +391,7 @@ EXAMPLES = r"""
     release_namespace: default
     force: True
     wait: True
+    wait_for_jobs: True
     replace: True
     update_repo_cache: True
     disable_hook: True
@@ -601,6 +608,7 @@ def deploy(
     release_values,
     chart_name,
     wait,
+    wait_for_jobs,
     wait_timeout,
     disable_hook,
     force,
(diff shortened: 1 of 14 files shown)


plugins/modules/helm.py
```diff
@@ -180,6 +180,12 @@
         It will wait for as long as I(wait_timeout). This feature requires helm>=3.7.0. Added in version 2.3.0.
     default: False
     type: bool
+  wait_for_jobs:
+    description:
+      - When I(release_state) is set to C(present), and I(wait) is set to C(True), wait until all jobs are in a
+        completed state before marking the release as successful. Added in version 6.5.0.
```
yurnov:
```suggestion
        completed state before marking the release as successful.
      - Is ignored when used without C(wait)
      - This parameters requires helm >= 3.5.0
    version_added: 6.5.0
```

plugins/modules/helm.py
```diff
@@ -622,6 +630,8 @@ def deploy(
 
     if wait:
         deploy_command += " --wait"
+        if wait_for_jobs:
+            deploy_command += " --wait-for-jobs"
```
yurnov:
```suggestion
            deploy_command += " --wait-for-jobs"
            if LooseVersion(helm_version) < LooseVersion("3.5.0"):
                module.fail_json(
                    msg="wait_for_jobs requires helm >= 3.5.0, current version is {0}".format(
                    helm_version
                    )
                )
            else:
                deploy_command += "  --wait-for-jobs"
```

yurnov (changes requested):
Please add a [changelog fragment ](https://docs.ansible.com/projects/ansible/latest/community/development_process.html#creating-a-changelog-fragment) and integration test for this change

plugins/modules/helm.py
abikouo:
```suggestion
        completed state before marking the release as successful.
      - Ignored when used without O(wait).
      - Requires Helm >= 3.5.0
   version_added: 6.6.0
```

abikouo (approved):

PR #741, opened 2024-06-01
Doc: add example of using kubectl connection plugin

yurnov:
##### SUMMARY

Currently documentation for collection don't include any examples of using `kubenrenes.core.kubectl` connection plugin and it's hard to start using that plugin. 

##### ISSUE TYPE
- Docs Pull Request

##### COMPONENT NAME
`kubenrenes.core.kubectl` connection plugin

##### ADDITIONAL INFORMATION
This PR was inspired by #288 and based on feedback on that PR and my own experience. Thanks @tpo for his try and @geerlingguy for his [Ansible for DevOps](https://www.ansiblefordevops.com/) book

PR #1196, opened 2026-07-14
Fix ansible-lint no-changed-when and execution-environment schema warnings

dbrennand:
## Summary

- Add `changed_when: false` to the three example `ansible.builtin.command` tasks in the `kubectl` connection plugin EXAMPLES block to resolve `no-changed-when` warnings from ansible-lint
- Add `version: 3` to `meta/execution-environment.yml` to satisfy the `schema[execution-environment]` rule which requires the `version` property (see [ansible-builder docs](https://ansible-builder.readthedocs.io/en/latest/definition/))

## Test plan

- [x] Verified with `ansible-lint 24.12.2` (ansible-core 2.16) that both `no-changed-when` and `schema[execution-environment]` warnings are resolved

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/changelogs/fragments/fix-ansible-lint-warnings.yaml
+++ b/changelogs/fragments/fix-ansible-lint-warnings.yaml
@@ -1,0 +1,3 @@
+trivial:
+  - kubectl connection plugin - added ``changed_when: false`` to example ``ansible.builtin.command`` tasks to resolve ansible-lint ``no-changed-when`` warnings.
+  - ee - added ``version: 3`` to ``meta/execution-environment.yml`` to resolve ansible-lint ``schema[execution-environment]`` warning.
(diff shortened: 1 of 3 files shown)


yurnov:
Here was discussion some time ago and it was deliberately not added `chanfe_when: false` to avoid confusion in the examples, please find details here https://github.com/ansible-collections/kubernetes.core/pull/741#issuecomment-2150515563

beeankha:
@yurnov thank you for the #741 context. I understand the concern about reader confusion from what was decided in that PR, but I think the changes here are required for Red Hat's collection validation protocol so I'm inclined to keep them as-is.

@dbrennand I think there might be a gap that needs to be addressed in this PR for the `var-naming[no-reserved]` warning for namespace in `kubectl.py` at line 46:

```
plugins/connection/kubectl.py:46: var-naming[no-reserved]: Variables names must not be Ansible reserved names. (namespace) (warning)
```
Do we maybe need a `skip` at the end of the [existing ignore entry](https://github.com/ansible-collections/kubernetes.core/blob/main/.ansible-lint-ignore#L5) for that?

yurnov:
Hi @beeankha,

Thanks for the answer, no objection from my side.

oraNod (approved):
Thanks @dbrennand

oraNod:
@beeankha @yurnov Any objections to merging this? I think everything has been addressed but wanted to check. Thanks.

yurnov:
> @beeankha @yurnov Any objections to merging this? I think everything has been addressed but wanted to check. Thanks.

No objections from my side

abikouo (approved):

oraNod:
This one is pretty straightforward so I'm going to proceed with the merge and backports. Thanks @dbrennand and everyone else for the reviews.

PR #1206, opened 2026-08-07
feat(helm): add cleanup_on_fail option

### LINKED ISSUE #1198: helm: add cleanup_on_fail parameter (maps to --cleanup-on-fail)
### Summary

`kubernetes.core.helm` exposes `atomic` (Helm `--atomic` / `--rollback-on-failure`) but has no equivalent for Helm's `--cleanup-on-fail`.

These are related but not the same:

| Flag | Behaviour |
|------|-----------|
| `--atomic` | Install fail → uninstall release; upgrade fail → rollback to last successful revision (implies wait) |
| `--cleanup-on-fail` | On failed upgrade, delete **new** resources created in that attempt |

Helm allows combining them (`--atomic --cleanup-on-fail`) so a rollback can also drop orphans created during the failed upgrade. The module can only set `atomic` today.

### Issue Type

- Feature Idea

### Component Name

helm

### Additional Information

Suggested API (mirroring the CLI):

```yaml
- name: Upgrade with rollback + orphan cleanup on failure
  kubernetes.core.helm:
    name: my-release
    chart_ref: myrepo/mychart
    release_namespace: myns
    wait: true
    atomic: true
    cleanup_on_fail: true   # NEW → helm --cleanup-on-fail
    timeout: 20m
```

Low priority: `atomic` alone covers the common failure mode. This is CLI parity / belt-and-suspenders for orphan GC when used together with `atomic`.

yurnov:
##### SUMMARY

Adds a `cleanup_on_fail` option to the `helm` module, mapping to Helm's `--cleanup-on-fail`.

`atomic` and `--cleanup-on-fail` are related but not equivalent: `atomic` rolls a failed upgrade back to the last successful revision, but resources the failed attempt created can survive that rollback. `--cleanup-on-fail` deletes them. Helm accepts both flags together, so the two combine into "roll back *and* drop the orphans" — which the module could not express before.

Two design decisions worth calling out:

**No version gate.** `--cleanup-on-fail` was ported to Helm v3 in [helm/helm#6480](https://github.com/helm/helm/pull/6480) (merged to `dev-v3` on 2019-09-24, ahead of the 3.0.0 release), so it is present in **helm 3.0.0** — which is already the module-wide minimum enforced by `validate_helm_version()`. A `LooseVersion` check like the ones guarding `plain_http` or `skip_schema_validation` would therefore be unreachable. Unlike `--atomic`, the flag was also *not* renamed in Helm v4, so no v3/v4 branch is needed either.

**Guarded against `replace`.** Helm only accepts `--cleanup-on-fail` on `upgrade` (and `rollback`), not on `install`. The module deploys through `helm install` when `replace=true`, so `cleanup_on_fail=true` combined with `replace=true` now fails with an explicit message instead of emitting a command Helm rejects.

This is implemented as a `fail_json()` inside `deploy()` rather than a `mutually_exclusive` entry in the argument spec, because Ansible evaluates mutual exclusion *before* defaults are applied — an argument-spec entry would spuriously reject an explicit `cleanup_on_fail: false` alongside `replace: true`. The existing `force_conflicts` / `server_side=false` guard uses the same approach.

Fixes #1198

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

helm

##### ADDITIONAL INFORMATION

Suggested usage, as requested in the issue:

```yaml
- name: Upgrade with rollback + orphan cleanup on failure
  kubernetes.core.helm:
    name: my-release
    chart_ref: myrepo/mychart
    release_namespace: myns
    wait: true
    atomic: true
    cleanup_on_fail: true
    timeout: 20m
```

Also included: a changelog fragment and the regenerated `docs/kubernetes.core.helm_module.rst` entry (`collection_prep_add_docs`; unrelated files the generator touched were reverted to keep the diff focused).

Six unit tests were added in `TestCleanupOnFail`, covering default-off, helm v3, helm v4 (same flag name), both `atomic` combinations, and the `replace` rejection.

The flag behaviour was verified against real Helm binaries rather than from documentation — Helm 3.0.0 via the `alpine/helm:3.0.0` image and Helm 4.2.2 locally.

### DIFF
--- a/tests/unit/modules/test_module_helm.py
+++ b/tests/unit/modules/test_module_helm.py
@@ -23,6 +23,48 @@ from ansible_collections.kubernetes.core.tests.unit.utils.ansible_module_mock im
     set_module_args,
 )
 
+CHART_INFO = {
+    "apiVersion": "v2",
+    "appVersion": "default",
+    "description": "A chart used in molecule tests",
+    "name": "test-chart",
+    "type": "application",
+    "version": "0.1.0",
+}
+
+
+def patch_ansible_module(test_case):
+    """Patch AnsibleModule's exit/fail/bin-path hooks for the duration of test_case."""
+    mock_module_helper = patch.multiple(
+        basic.AnsibleModule,
+        exit_json=exit_json,
+        fail_json=fail_json,
+        get_bin_path=get_bin_path,
+    )
+    mock_module_helper.start()
+    test_case.addCleanup(mock_module_helper.stop)
+
+
+def deploy_command_for_version(args, helm_version):
+    """Run helm.main() against a mocked helm of helm_version and return its command.
+
+    Only the fresh-install path is mocked, so the caller sees the command built by
+    deploy(). Failures raised by the module (AnsibleFailJson) propagate to the test.
+    """
+    set_module_args(args)
+    helm.get_release_status = MagicMock(return_value=None)
+    helm.fetch_chart_info = MagicMock(return_value=CHART_INFO)
+    with patch.object(
+        helm.AnsibleHelmModule, "get_helm_version", return_value=helm_version
+    ):
+        with patch.object(basic.AnsibleModule, "run_command") as mock_run_command:
+            mock_run_command.return_value = (0, "configuration updated", "")
+            try:
+                helm.main()
+            except AnsibleExitJson as exc:
+                return exc.args[0]["command"]
+    raise AssertionError("helm.main() did not exit successfully")
+
 
 class TestDependencyUpdateWithoutChartRepoUrlOption(unittest.TestCase):
     def setUp(self):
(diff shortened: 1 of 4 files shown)


yurnov:
Refactored tests to decrease code duplication

tests/unit/modules/test_module_helm.py
```diff
@@ -788,3 +801,90 @@ def test_atomic_with_replace_uses_rollback_on_failure_on_helm_v4(self):
...
+            {
+                "release_name": "test",
+                "release_namespace": "test",
+                "chart_ref": "chart1",
+                "atomic": True,
+                "cleanup_on_fail": True,
+            },
+            "4.0.0",
+        )
+        assert "--rollback-on-failure" in command
```
abikouo:
```suggestion
        assert "--rollback-on-failure" in command
        assert "--atomic" not in command
```

abikouo (approved):
LGTM, just a minor remark.
Thanks @yurnov

tests/unit/modules/test_module_helm.py
```diff
@@ -788,3 +801,90 @@ def test_atomic_with_replace_uses_rollback_on_failure_on_helm_v4(self):
...
+            {
+                "release_name": "test",
+                "release_namespace": "test",
+                "chart_ref": "chart1",
+                "atomic": True,
+                "cleanup_on_fail": True,
+            },
+            "3.17.0",
+        )
+        assert "--atomic" in command
```
abikouo, line 861:
```suggestion
        assert "--atomic" in command
        assert "--rollback-on-failure" not in command
```

tests/unit/modules/test_module_helm.py
yurnov: Valid, will add as the new commit

tests/unit/modules/test_module_helm.py
yurnov: valid

yurnov:
> LGTM, just a minor remark. Thanks @yurnov

Thanks @abikouo, pushed your suggestion, please re-review and remove `needs_triage` tag (instead, `backport-6` is recomended)

beeankha (approved):

PR #1197, opened 2026-07-23
Remove remaining `ansible.module_utils.six` import from kubectl connection plugin

ChihweiLHBird:
##### SUMMARY

#998 removed the `ansible.module_utils.six` imports, but `plugins/connection/kubectl.py` was missed and still imports the shim. This replaces it with the stdlib equivalent:

```python
-from ansible.module_utils.six.moves import shlex_quote
+from shlex import quote as shlex_quote
```

On Python 3 `six.moves.shlex_quote` is already `shlex.quote` (the same function object), so there is no behaviour change at the single call site in `put_file()`. The import is placed in the stdlib group to keep `tox -e linters` clean.

`ansible.module_utils.six` is deprecated in ansible-core 2.21 and scheduled for removal in 2.24.

##### ISSUE TYPE

- Refactoring Pull Request

##### COMPONENT NAME

plugins/connection/kubectl.py

PR #1209, opened 2026-08-08
fix: avoid deprecated `Templar.environment` access when rendering `template`

### LINKED ISSUE #1208: k8s/k8s_info: deprecated `Templar.environment` access when using `template` (breaks on ansible-core 2.23)
##### SUMMARY

The `k8s` / `k8s_info` action plugin reads the deprecated `Templar.environment` property whenever the `template` parameter is used, so every templated task emits:

```
[DEPRECATION WARNING]: Direct access to the `environment` attribute is deprecated. This feature will be removed from ansible-core version 2.23. Consider using `copy_with_new_env` or passing `overrides` to `template`.
```

ansible-core deprecated the property in 2.19 and schedules removal in **2.23**, so this is not only noise — `kubernetes.core.k8s` with `template:` will break outright once the property is gone.

The access is in [`plugins/action/k8s_info.py`](https://github.com/ansible-collections/kubernetes.core/blob/main/plugins/action/k8s_info.py) (`plugins/action/k8s.py` is a symlink to it), inside `get_template_args` handling:

- L233-L234 — building `default_environment`; correctly guarded by `if trust_as_template is None`, so it does not fire on modern core.
- **L253** — `if hasattr(self._templar.environment, key)`; **not** guarded, so it runs on every core version.
- L257, L260 — `setattr(...)`; guarded by `if trust_as_template is None`.

L253 is the one that fires on current core. It probes 7 attribute names (`newline_sequence`, `variable_start_string`, `variable_end_string`, `block_start_string`, `block_end_string`, `trim_blocks`, `lstrip_blocks`) per template file, and each probe triggers the property. On a repo with ~58 playbooks using `template:` this drowns the deploy output.

On the `trust_as_template` path the result of that `hasattr` is only used to decide whether a key belongs in `overrides` — which is already passed to `self._templar.template(...)`. So the property access is not needed there at all; a static tuple of the known Jinja environment option names would do, or `copy_with_new_env`.

##### Prior art in other collections

Several collections have already made this migration and are useful references:

- **`ansible-collections/ansible.windows`** — [`plugins/action/win_template.py#L166-L201`](https://github.com/ansible-collections/ansible.windows/blob/main/plugins/action/win_template.py#L166-L201) is the closest analogue, since this code is essentially a fork of the core `template` action plugin. It builds `overrides` directly from the task arguments (no environment probing) and uses `self._templar.copy_with_new_env(...)`, branching on a `USE_DATA_TAGGING` flag for the pre/post-2.19 paths.
- **`aristanetworks/avd`** — issue [#6649](https://github.com/aristanetworks/avd/issues/6649) tracked this exact warning; fixed in PR [#6658](https://github.com/aristanetworks/avd/pull/6658) ("Add AVDTemplar to avoid accessing private Ansible Templar attributes and triggering deprecation"), merged 2026-03-09.
- **`ansible-collections/community.general`** — `plugins/lookup/dependent.py` and `plugins/lookup/merge_variables.py` now use `copy_with_new_env`.

I could not find an existing issue for this in this repo; happy to open a PR following the `ansible.windows` pattern if that direction sounds right.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

`k8s`, `k8s_info` (shared action plugin `plugins/action/k8s_info.py`)

##### ANSIBLE VERSION

```
ansible [core 2.21.2]
  config file = None
  configured module search path = ['~/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = .venv/lib/python3.14/site-packages/ansible
  ansible collection location = ~/.ansible/collections:/usr/share/ansible/collections
  executable location = .venv/bin/ansible
  python version = 3.14.7 (main, Aug  5 2026, 15:35:18) [Clang 22.1.3]
  jinja version = 3.1.6
  pyyaml version = 6.0.3 (with libyaml v0.2.5)
```

##### COLLECTION VERSION

```
# ~/.ansible/collections/ansible_collections
Collection      Version
--------------- -------
kubernetes.core 6.5.0
```

`main` is also affected — the L253 access is unchanged there.

##### CONFIGURATION

```
CONFIG_FILE() = None

GALAXY_SERVERS:
```

##### OS / ENVIRONMENT

macOS 15 (Darwin 25.5.0), arm64, Python 3.14.7.

##### STEPS TO REPRODUCE

No reachable cluster is required — the warning is emitted by the action plugin before the module ever connects.

`manifest/namespace.yaml`:

```yaml
---
apiVersion: v1
kind: Namespace
metadata:
  name: "{{ ns_name }}"
```

`repro.yaml`:

```yaml
---
- name: Reproduce templar.environment deprecation
  hosts: localhost
  connection: local
  gather_facts: false

  vars:
    ns_name: repro

  tasks:
    - name: Apply a templated manifest
      kubernetes.core.k8s:
        state: present
        template: manifest/namespace.yaml
```

```
ansible-playbook -i localhost, repro.yaml
```

##### EXPECTED RESULTS

The manifest is templated and applied with no deprecation warning, and the plugin keeps working on ansible-core 2.23+.

##### ACTUAL RESULTS

```
PLAY [Reproduce templar.environment deprecation] *******************************

TASK [Apply a templated manifest] **********************************************
[WARNING]: Deprecation warnings can be disabled by setting `deprecation_warnings=False` in ansible.cfg.
[DEPRECATION WARNING]: Direct access to the `environment` attribute is deprecated. This feature will be removed from ansible-core version 2.23. Consider using `copy_with_new_env` or passing `overrides` to `template`.
[ERROR]: Task failed: Module failed: HTTPSConnectionPool(host='127.0.0.1', port=1): Max retries exceeded with url: /version (Caused by NewConnectionError("HTTPSConnection(host='127.0.0.1', port=1): Failed to establish a new connection: [Errno 61] Connection refused"))

PLAY RECAP *********************************************************************
localhost                  : ok=0    changed=0    unreachable=0    failed=1    skipped=0    rescued=0    ignored=0
```

(The connection error is just the dummy kubeconfig used to keep the reproducer self-contained; the deprecation warning precedes it.)

jheddings:
##### SUMMARY

Fixes #1208.

`load_template()` probes `self._templar.environment` to decide which keys of `template_item` are Jinja2 environment options. Direct access to that property is deprecated in ansible-core 2.19 and removed in **2.23**, so every task using `template:` emits a deprecation warning today and will fail outright on 2.23.

The probed keys are the exact set `get_template_args()` builds a few lines earlier, so the environment is not actually needed — this replaces the probe with a static tuple:

```python
-                    if hasattr(self._templar.environment, key):
+                    if key in TEMPLATE_ENVIRONMENT_OPTIONS:
```

The same tuple also replaces the inline literal in the `trust_as_template is None` branch above. That branch (and the `setattr` calls) only runs on pre-2.19 core where the property is not deprecated, so it is left otherwise untouched.

No behaviour change: on any supported Jinja2 these seven names are all real `Environment` attributes, and `path` — the only other key — is not. `overrides` is already passed to `Templar.template()`, so the modern path never needed the environment at all.

##### Verification

Rendering was compared before and after by driving `load_template()` against a real `Templar` across five cases — plain string arg, dict arg with default delimiters, dict arg with custom `variable_start_string`/`variable_end_string`, dict arg with `trim_blocks`/`lstrip_blocks`, and a list arg. Output is byte-identical, and the custom-delimiter case still resolves `[[ ns_name ]]` while leaving `{{ ns_name }}` literal. The deprecation warning is gone from stderr.

Existing integration coverage for this path is `tests/integration/targets/k8s_template`, which exercises custom delimiters at `tasks/main.yml:115` and `:151`. No new tests are added since behaviour is unchanged; happy to add some if you would prefer.

`tox -e linters` equivalents (black, isort --profile black, flake8, yamllint) and `antsibull-changelog lint` are clean.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

plugins/action/k8s_info.py

##### ADDITIONAL INFORMATION

I filed the changelog fragment under `minor_changes` to match the precedent of #1197 (removing deprecated usage to silence warnings). Glad to move it to `bugfixes` if you would rather emphasise the 2.23 breakage.

Reproducer, needing no reachable cluster — the warning is emitted by the action plugin before the module connects:

```yaml
- hosts: localhost
  connection: local
  gather_facts: false
  vars:
    ns_name: repro
  tasks:
    - kubernetes.core.k8s:
        state: present
        template: manifest/namespace.yaml
```

Before:

```
[DEPRECATION WARNING]: Direct access to the `environment` attribute is deprecated. This feature will be removed from ansible-core version 2.23. Consider using `copy_with_new_env` or passing `overrides` to `template`.
```

After: no warning.

### DIFF
--- a/changelogs/fragments/20260808-k8s-templar-environment-deprecation.yaml
+++ b/changelogs/fragments/20260808-k8s-templar-environment-deprecation.yaml
@@ -1,0 +1,5 @@
+---
+minor_changes:
+  - k8s - Avoid direct access to ``Templar.environment`` when rendering the ``template``
+    parameter, which is deprecated in ansible-core 2.19 and removed in 2.23
+    (https://github.com/ansible-collections/kubernetes.core/issues/1208).
--- a/changelogs/fragments/20260808-k8s-templar-environment-deprecation.yaml
+++ b/changelogs/fragments/20260808-k8s-templar-environment-deprecation.yaml
@@ -1,5 +1,8 @@
 ---
 minor_changes:
   - k8s - Avoid direct access to ``Templar.environment`` when rendering the ``template``
-    parameter, which is deprecated in ansible-core 2.19 and removed in 2.23
+    parameter, which is deprecated in ansible-core 2.19 and removed in 2.23. Jinja2
+    environment options such as ``variable_start_string`` are now applied as templating
+    overrides on every supported ansible-core version, so custom delimiters behave
+    consistently across versions
     (https://github.com/ansible-collections/kubernetes.core/issues/1208).
--- a/plugins/action/k8s_info.py
+++ b/plugins/action/k8s_info.py
@@ -44,6 +44,19 @@ def _from_yaml_to_definition(buffer):
 
 ENV_KUBECONFIG_PATH_SEPARATOR = ";" if platform.system() == "Windows" else ":"
 
+# Jinja2 environment options accepted by the 'template' parameter. Kept as a static
+# tuple rather than probing Templar.environment, direct access to which is deprecated
+# in ansible-core 2.19 and removed in 2.23.
+TEMPLATE_ENVIRONMENT_OPTIONS = (
+    "newline_sequence",
+    "variable_start_string",
+    "variable_end_string",
+    "block_start_string",
+    "block_end_string",
+    "trim_blocks",
+    "lstrip_blocks",
+)
+
 
 class ActionModule(ActionBase):
     TRANSFERS_FILES = True
@@ -221,15 +234,7 @@ class ActionModule(ActionBase):
 
         default_environment = {}
         if trust_as_template is None:
-            for key in (
-                "newline_sequence",
-                "variable_start_string",
-                "variable_end_string",
-                "block_start_string",
-                "block_end_string",
-                "trim_blocks",
-                "lstrip_blocks",
-            ):
+            for key in TEMPLATE_ENVIRONMENT_OPTIONS:
                 if hasattr(self._templar.environment, key):
                     default_environment[key] = getattr(self._templar.environment, key)
         for template_item in template_params:
@@ -250,7 +255,7 @@ class ActionModule(ActionBase):
                 temp_vars = copy.deepcopy(task_vars)
                 overrides = {}
                 for key, value in template_item.items():
-                    if hasattr(self._templar.environment, key):
+                    if key in TEMPLATE_ENVIRONMENT_OPTIONS:
                         if value is not None:
                             overrides[key] = value
                             if trust_as_template is None:
--- a/plugins/action/k8s_info.py
+++ b/plugins/action/k8s_info.py
@@ -44,9 +44,9 @@ def _from_yaml_to_definition(buffer):
 
 ENV_KUBECONFIG_PATH_SEPARATOR = ";" if platform.system() == "Windows" else ":"
 
-# Jinja2 environment options accepted by the 'template' parameter. Kept as a static
-# tuple rather than probing Templar.environment, direct access to which is deprecated
-# in ansible-core 2.19 and removed in 2.23.
+# Jinja2 environment options accepted by the 'template' parameter. These are passed
+# to the templar as overrides; Templar.environment is not touched directly, as that
+# is deprecated in ansible-core 2.19 and removed in 2.23.
 TEMPLATE_ENVIRONMENT_OPTIONS = (
     "newline_sequence",
     "variable_start_string",
@@ -232,11 +232,6 @@ class ActionModule(ActionBase):
         result_template = []
         old_vars = self._templar.available_variables
 
-        default_environment = {}
-        if trust_as_template is None:
-            for key in TEMPLATE_ENVIRONMENT_OPTIONS:
-                if hasattr(self._templar.environment, key):
-                    default_environment[key] = getattr(self._templar.environment, key)
         for template_item in template_params:
             # We need to convert unescaped sequences to proper escaped sequences for Jinja2
             newline_sequence = template_item["newline_sequence"]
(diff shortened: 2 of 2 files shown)


abikouo (changes requested):
We need to clean up all remaining references to `self._templar.environment`. The code already uses the overrides option for custom configurations during `template()` calls.

jheddings:
Thanks — done in c91e400. There are now no references to `self._templar.environment` left in `plugins/`.

The pre-2.19 branch was mutating the shared environment with `setattr` and capturing the prior values into `default_environment` so it could restore them afterwards. As you say, that is unnecessary: `Templar.do_template()` has accepted `overrides` since 2.16 (the collection's `requires_ansible` floor), where `_create_overlay()` applies them via `jinja_env.overlay(**overrides)` and `myenv.newline_sequence` is used for trailing-newline preservation. So both branches now build one `overrides` dict and neither touches the environment. `default_environment` and the version-conditional `setattr` calls are gone, which also removes the `trust_as_template is None` special-casing from that block entirely.

##### One behaviour change worth flagging

There is a single user-visible difference, and only on ansible-core < 2.19. `do_template()` carries this note:

> Creating an overlay that lives only inside do_template means that overrides are not applied when templating nested variables in AnsibleJ2Vars where Templar.environment is used, not the overlay.

Because the old code mutated the environment globally, custom delimiters also reached *nested* variable renders. With `overrides` they do not. Rendering a manifest containing `[[ ns_name ]]` with `variable_start_string: '[['`, where `ns_name`'s own value is `"[[ inner ]]"`:

| | core 2.16 | core 2.21 |
|---|---|---|
| before | `resolved-nested` | `[[ inner ]]` |
| after | `[[ inner ]]` | `[[ inner ]]` |

So this aligns old cores with what 2.19+ already does rather than introducing new behaviour — the same playbook previously rendered differently depending on the controller's core version. It is also not avoidable, since the mutation API is removed in 2.23. I noted the consistency improvement in the changelog fragment; happy to split it into a separate `bugfixes` entry if you would rather call it out on its own.

##### Verification

`load_template()` was driven against a real `Templar` on **ansible-core 2.16.19** and **2.21.2**, across five argument shapes — plain string, dict with default delimiters, dict with custom `variable_start_string`/`variable_end_string`, dict with `trim_blocks`/`lstrip_blocks`, and a list. Output is byte-identical to the current `main` on both versions, apart from the nested case above. The deprecation warning is gone from 2.21.

black, `isort --profile black`, flake8, `yamllint -s .` and `antsibull-changelog lint` are clean.

beeankha (approved):

PR #992, opened 2025-08-15
Extend k8s action group

jplitza:
##### SUMMARY

Add all k8s_* modules to the action group in order to esaily set kubeconfig parameter

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

meta

##### ADDITIONAL INFORMATION

```paste below

```

PR #1216, opened 2026-08-12
fix(meta): add missing modules to helm and k8s action groups, document them

### LINKED ISSUE #577: Add documentation on action groups
##### SUMMARY

The kubernetes.core collection has two action groups, kubernetes.core.k8s and kubernetes.core.helm, but their existence and usage is not documented anywhere. This should probably be added to the scenario guide. This can be especially confusing given that there is the legacy k8s module defaults group.

##### ISSUE TYPE
- Documentation Report

##### ANSIBLE VERSION

```paste below

```

yurnov:
##### SUMMARY

Three modules build their argument spec from the shared authentication arg specs but were
never listed in the corresponding action group in `meta/runtime.yml`:

| Module | Builds its spec from | Missing from group |
| --- | --- | --- |
| `helm_plugin` | `HELM_AUTH_ARG_SPEC` | `helm` |
| `helm_plugin_info` | `HELM_AUTH_ARG_SPEC` | `helm` |
| `k8s_taint` | `AUTH_ARG_SPEC` | `k8s` |

Because `module_defaults` is applied per group member, anyone setting defaults on
`group/kubernetes.core.helm` or `group/kubernetes.core.k8s` silently got nothing for these
three. Nothing warns — the task just runs without the defaults the user set, which for
options like `kubeconfig`, `context` or `api_key` means quietly talking to the wrong cluster
rather than failing. See the reproduction below.

This also documents both action groups in the scenario guide, which is what #577 asked for.
The new section covers what the groups contain, a worked `module_defaults` example, the
distinction from the legacy `k8s` group that ansible-core defines (the issue called this out
specifically as a source of confusion), and which modules are deliberately *not* in the
groups.

On that last point — these four stay out on purpose, because they do not accept the shared
authentication options and adding them would make `module_defaults` fail with
`Unsupported parameters` for everyone using the group:

- `helm_pull` and `helm_template` accept only `binary_path`.
- `helm_registry_auth` accepts `binary_path` plus its own required `host`, which is an OCI
  registry URL rather than a Kubernetes API server.
- `kubeconfig` writes a kubeconfig file locally and takes no cluster connection options.

Finally, this adds a unit test that pins group membership in both directions, since the
groups have drifted before (PR #992 fixed the same class of omission for `k8s_cluster_info`,
`k8s_json_patch` and `k8s_rollback`) and nothing prevented it from happening again.

Fixes #577

##### ISSUE TYPE

- Bugfix Pull Request
- Docs Pull Request

##### COMPONENT NAME

`meta/runtime.yml`, `helm_plugin`, `helm_plugin_info`, `k8s_taint`, docsite scenario guide

##### ADDITIONAL INFORMATION

**Membership rule.** A module belongs in an action group if and only if it accepts that
group's shared authentication options. Both directions are bugs: a module that inherits the
shared spec but is absent from the group silently ignores `module_defaults`, and a module
listed in the group that does not accept the options makes `module_defaults` fail for every
user of that group.

**Test.** `tests/unit/test_action_groups.py` asserts that invariant both ways, that every
listed member exists under `plugins/modules/`, and that the four excluded modules stay
excluded (with their reasons recorded, so restoring one has to be a deliberate decision
rather than an oversight). Membership is read statically with `ast`, so the test needs no
imports, no cluster and no `kubernetes` package, and runs in well under a second.

I verified the test actually fails when it should, rather than only passing:

| Injected fault | Result |
| --- | --- |
| Revert `meta/runtime.yml` to its pre-fix state | fails, naming `helm_plugin`, `helm_plugin_info`, `k8s_taint` |
| Add `helm_template` to the `helm` group | fails, warning it would break `module_defaults` |
| List a nonexistent module in a group | fails on both the existence and the spec check |

**Test results.** Full unit suite: 267 passed (258 before this change, plus the 9 new cases).
`black`, `isort`, `flake8`, `yamllint` and `antsibull-changelog lint` are all clean.

Reproduction below uses `validate_certs: not-a-boolean` as an observable marker: if the group
default reaches the module, argument-spec validation rejects it; if it does not, the module
runs as though no default were set.

```paste below
$ cat demo.yml
- hosts: localhost
  connection: local
  gather_facts: false
  module_defaults:
    group/kubernetes.core.k8s:
      validate_certs: not-a-boolean
  tasks:
    - name: Taint a node, relying on module_defaults from the k8s action group
      kubernetes.core.k8s_taint:
        name: node1
        taints:
          - key: dedicated
            effect: NoSchedule

# BEFORE — k8s_taint absent from the k8s action group.
# The group default is silently dropped, so the module runs without it and
# gets all the way to opening the kubeconfig:

$ ansible-playbook -i localhost, demo.yml
fatal: [localhost]: FAILED! => {"changed": false, "msg": "Could not create API client:
Invalid kube-config file. Expected key current-context in /home/user/.kube/config"}

# AFTER — k8s_taint listed in the k8s action group.
# The group default now reaches the module and is validated:

$ ansible-playbook -i localhost, demo.yml
fatal: [localhost]: FAILED! => {"changed": false, "msg": "argument 'validate_certs' is of
type <class 'str'> and we were unable to convert to bool: The value 'not-a-boolean' is not
a valid boolean."}
```

### DIFF
--- a/changelogs/fragments/20260812-action-groups.yaml
+++ b/changelogs/fragments/20260812-action-groups.yaml
@@ -1,6 +1,3 @@
 ---
 minor_changes:
-  - docs - Document the ``kubernetes.core.k8s`` and ``kubernetes.core.helm`` action groups in the scenario guide, including which modules are deliberately excluded from them and how they differ from the legacy ``k8s`` group defined by ansible-core (https://github.com/ansible-collections/kubernetes.core/pull/1216).
-
-bugfixes:
   - meta - Add ``helm_plugin`` and ``helm_plugin_info`` to the ``helm`` action group, and ``k8s_taint`` to the ``k8s`` action group, so that ``module_defaults`` set on those groups applies to them (https://github.com/ansible-collections/kubernetes.core/pull/1216).
--- a/changelogs/fragments/20260812-action-groups.yaml
+++ b/changelogs/fragments/20260812-action-groups.yaml
@@ -1,0 +1,12 @@
+---
+minor_changes:
+  - docs - Document the ``kubernetes.core.k8s`` and ``kubernetes.core.helm`` action groups in
+    the scenario guide, including which modules are deliberately excluded from them and how
+    they differ from the legacy ``k8s`` group defined by ansible-core
+    (https://github.com/ansible-collections/kubernetes.core/pull/1216).
+
+bugfixes:
+  - meta - Add ``helm_plugin`` and ``helm_plugin_info`` to the ``helm`` action group, and
+    ``k8s_taint`` to the ``k8s`` action group, so that ``module_defaults`` set on those groups
+    applies to them
+    (https://github.com/ansible-collections/kubernetes.core/pull/1216).
--- a/changelogs/fragments/20260812-action-groups.yaml
+++ b/changelogs/fragments/20260812-action-groups.yaml
@@ -1,12 +1,6 @@
 ---
 minor_changes:
-  - docs - Document the ``kubernetes.core.k8s`` and ``kubernetes.core.helm`` action groups in
-    the scenario guide, including which modules are deliberately excluded from them and how
-    they differ from the legacy ``k8s`` group defined by ansible-core
-    (https://github.com/ansible-collections/kubernetes.core/pull/1216).
+  - docs - Document the ``kubernetes.core.k8s`` and ``kubernetes.core.helm`` action groups in the scenario guide, including which modules are deliberately excluded from them and how they differ from the legacy ``k8s`` group defined by ansible-core (https://github.com/ansible-collections/kubernetes.core/pull/1216).
 
 bugfixes:
-  - meta - Add ``helm_plugin`` and ``helm_plugin_info`` to the ``helm`` action group, and
-    ``k8s_taint`` to the ``k8s`` action group, so that ``module_defaults`` set on those groups
-    applies to them
-    (https://github.com/ansible-collections/kubernetes.core/pull/1216).
+  - meta - Add ``helm_plugin`` and ``helm_plugin_info`` to the ``helm`` action group, and ``k8s_taint`` to the ``k8s`` action group, so that ``module_defaults`` set on those groups applies to them (https://github.com/ansible-collections/kubernetes.core/pull/1216).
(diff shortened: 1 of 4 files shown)


beeankha (approved):

changelogs/fragments/20260812-action-groups.yaml
```diff
@@ -0,0 +1,6 @@
+---
+minor_changes:
```
abikouo, line 2: Documentation updates do not require a changelog.

changelogs/fragments/20260812-action-groups.yaml
```diff
@@ -0,0 +1,6 @@
...
+  - docs - Document the ``kubernetes.core.k8s`` and ``kubernetes.core.helm`` action groups in the scenario guide, including which modules are deliberately excluded from them and how they differ from the legacy ``k8s`` group defined by ansible-core (https://github.com/ansible-collections/kubernetes.core/pull/1216).
+
+bugfixes:
```
abikouo:
```suggestion
minor_changes:
```
This is not considered as a bugfix since there is no place where the collection was stating that these modules are part of the action group

docs/docsite/rst/kubernetes_scenarios/k8s_intro.rst
```diff
@@ -42,6 +42,71 @@ Basic authentication is also supported using the ``username`` and ``password`` o
...
+
+Prefix the group name with ``group/`` to use it:
+
+.. code-block:: yaml
+
+    - hosts: localhost
+      module_defaults:
+        group/kubernetes.core.k8s:
+          kubeconfig: /path/to/kubeconfig
+          context: staging
```
abikouo, line 73:
```suggestion
        group/kubernetes.core.k8s:
          kubeconfig: /path/to/kubeconfig
          context: staging
        group/kubernetes.core.helm:
           binary_path: /path/to/helm/binary
           context: staging
           kubeconfig: /path/to/kubeconfig
```
And add some Helm tasks that use these parameters

abikouo (changes requested):
Looks good overall, thanks @yurnov.
Just some minor remarks

changelogs/fragments/20260812-action-groups.yaml
yurnov: Done

changelogs/fragments/20260812-action-groups.yaml
yurnov: Done

abikouo (approved):

Issue #621, opened 2023-05-18
kubernetes.core.k8s_cp freezes


raman-babich:
##### SUMMARY

I am trying to run kubernetes.core.k8s_cp task in my role for copy dir(25Mb) from pod but it stucks on execution of this task.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

kubernetes.core.k8s_cp

##### ANSIBLE VERSION

I am using HIDDEN for possibly sensetive info.
```paste below
ansible [core 2.14.2]
  config file = None
  configured module search path = ['/Users/<HIDDEN>/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /opt/homebrew/Cellar/ansible/7.2.0/libexec/lib/python3.11/site-packages/ansible
  ansible collection location = /Users/<HIDDEN>/.ansible/collections:/usr/share/ansible/collections
  executable location = /opt/homebrew/bin/ansible
  python version = 3.11.2 (main, Feb 13 2023, 03:31:04) [Clang 14.0.0 (clang-1400.0.29.202)] (/opt/homebrew/Cellar/ansible/7.2.0/libexec/bin/python3.11)
  jinja version = 3.1.2
  libyaml = True
```

##### COLLECTION VERSION

```paste below
Collection        Version
----------------- -------
community.general 6.3.0  
```
Attaching `pip list` also maybe it could help.
```
Package            Version
------------------ --------
cachetools         5.3.0
certifi            2023.5.7
charset-normalizer 3.1.0
docker             5.0.3
google-auth        2.18.1
idna               3.4
kubernetes         24.2.0
oauthlib           3.2.2
pip                23.1.2
pyasn1             0.5.0
pyasn1-modules     0.3.0
python-dateutil    2.8.2
PyYAML             6.0
requests           2.30.0
requests-oauthlib  1.3.1
rsa                4.9
setuptools         65.6.3
six                1.16.0
urllib3            1.26.15
websocket-client   1.5.1
wheel              0.38.4

```

##### CONFIGURATION

```paste below
CONFIG_FILE() = None
```

##### OS / ENVIRONMENT

MacOs Ventura 13.1

##### STEPS TO REPRODUCE

```yaml
- name: 'Copy a file or a dir from a pod to local file or dir'
  kubernetes.core.k8s_cp:
    kubeconfig: '{{ copy_from_pod_kubeconfig }}'
[...]

PR #1217, opened 2026-08-13
k8s_cp - wait for remote tar to finish before closing the stream

### LINKED ISSUE #776: kubernetes.core.k8s_cp intermittently copying incomplete files to pod
##### SUMMARY
Occasionally I've noticed that when using the kubernetes.core.k8s_cp module the file that gets copied is not complete.  This isn't happening all the time but it is causing issues for us.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
kubernetes.core.k8s_cp

##### ANSIBLE VERSION
quay.io/ansible/ansible-builder:latest

##### STEPS TO REPRODUCE
```
 - name: Copy metricbeat index template to elasticsearch master pod
   kubernetes.core.k8s_cp:
     kubeconfig: "{{ context_file }}"
     context: "{{ aks_name }}"
     namespace: "{{ namespace }}"
     pod: "{{ prefix_type }}-elasticsearch-master-0"
     container: "{{ prefix_type }}-elasticsearch"
     local_path: "{{ role_path }}/files/metricbeat_template.json"
     remote_path: /tmp/metricbeat_template.json
   no_log: "{{ no_log }}"
```

##### EXPECTED RESULTS
file to be copied in it's entirety

##### ACTUAL RESULTS
source file:
```
wc -l metricbeat_template.json
35766 metricbeat_template.json
```

file on pod after running above ansible task:
```
wc -l /tmp/metricbeat_template.json
35225 /tmp/metricbeat_template.json
```

this doesn't happen every time but I've caught this happening twice now.

re-running the playbook results in the file being copied in it's entirety (from the pod):
```
wc -l /tmp/metricbeat_template.json
35766 /tmp/metricbeat_template.json
```

yurnov:
##### SUMMARY

Fixes #776.

`K8SCopyToPod.run()` breaks out of its write loop one iteration after the last chunk and closes the exec connection right away. `write_stdin()` only hands bytes to the local socket. They still have to cross the API server and reach the container, so closing at that point kills `tar` mid-extract and leaves a truncated file. The exit status was never checked either, so the partial copy was still reported as `changed=true`. That matches the report: a short file on the first run, complete on retry.

`_run_from_pod()` in the same file already polls `while resp.is_open()` and checks the status on the error channel. The write path never did.

Waiting for `tar` to exit is not enough on its own. The v4 websocket exec protocol cannot half-close stdin, so the only EOF the remote process can get is the connection going away. Relying on the end-of-archive marker that `tarfile` writes is not portable either. GNU tar exits on it, but busybox tar blocks for EOF regardless, even with extra zero records appended. I checked this against busybox 1.35. The integration pods run `busybox:latest`, so that approach would have turned every copy into a timeout. 

The archive size is known once the archive has been built, so this bounds the stream with `head -c N | tar ...` instead. `head` reads exactly N bytes and closes the pipe, `tar` gets a clean EOF and exits on its own, and the API server reports
its exit status on the error channel. The module now waits for that and checks it. No `close_channel()` or v5 support is needed, so the `kubernetes>=24.2.0` floor is unchanged. Every `WSClient` method used here has the same signature in 24.2.0 and 36.0.3.

Containers without `/bin/sh` and `head` keep the previous behaviour and get a warning that completion could not be verified, so `tar`-only images do not break.

Two smaller changes in the same code:

* `copy_timeout` (default 300s) bounds the write and the drain. The drain needs a deadline anyway, otherwise it becomes a new place to hang forever. This should also cover #621, but I could not reproduce that report, so I have not marked it as fixed.
* The archive is streamed from the temp file rather than pre-buffered into a list of 1 MiB chunks. The old code held the whole archive in memory, which is 200 MB for the large-file integration test.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME

k8s_cp

##### ADDITIONAL INFORMATION

The truncation is timing dependent and I could not reproduce it on a local kind cluster. Eight 64 MB copies over loopback all came through intact, which is consistent with the report coming from AKS over a real network.

The detection half of the bug is deterministic. With `readOnlyRootFilesystem: true` the remote `tar` has to fail, and before this change the module reported success anyway.

Before, on `main`:

```paste below
TASK [copy into a read-only filesystem (tar MUST fail)] ************************
changed: [localhost]

TASK [debug] *******************************************************************
    "msg": "module said: failed=False | changed=True | msg=/tmp/ro_payload.txt successfully copied into remote Pod into /ro_payload.txt"

TASK [debug] *******************************************************************
    "msg": "file on pod: ABSENT"
```

After:

```paste below
TASK [copy into a read-only filesystem (tar MUST fail)] ************************
fatal: [localhost]: FAILED! => {"changed": false, "command": ["/bin/sh", "-c", "head -c 10240 | tar -xmf - -C /"], "msg": "Failed to copy local file/directory into Pod: tar: can't remove old file ro_payload.txt: Read-only file system"}

TASK [debug] *******************************************************************
    "msg": "file on pod: ABSENT"
```

For the fallback path I deleted `/usr/bin/head` from a `debian:stable-slim` pod. The probe detects it, the warning fires, and the copy still completes with a matching checksum:

```paste below
    "msg": "head present? False (rc=127)"
    "msg": "cp.warnings=[\"Container has no '/bin/sh' with 'head', falling back to writing the archive straight to tar. The module cannot confirm the copy completed; verify the file after copying.\"]"
    "msg": "fallback copy intact: a12f0ffac620e2a37dab45122e6f80e5"
```

Testing, against kind with Kubernetes 1.36.1 and `kubernetes` 36.0.3:

* Full `k8s_copy` integration target: 179 ok, 0 failed. That includes the existing 200 MB binary round-trip and the busybox, multi-container, init-container, directory, space-in-name and check-mode cases.
* New unit tests in `tests/unit/module_utils/test_copy.py`. There was no unit coverage for `copy.py` before. They cover: the drain waits for exit, a connection dropped mid-archive fails instead of reporting success, a non-zero exit is surfaced, stderr written after the last chunk is still collected, the timeout is honoured, and the bounded and fallback command construction. Full suite: 268 passed.
* `ansible-test sanity` (`validate-modules`, `pep8`, `pylint`, `yamllint`) clean on all changed files.
* New integration test copies the same 12 MB file 10 times and compares each time, since a single copy rarely loses the race.

Note: this PR was prepared with AI assistance (Claude Code). The diagnosis, the busybox behaviour check and all test output above come from real runs, not from generated text.

### DIFF
--- a/plugins/module_utils/copy.py
+++ b/plugins/module_utils/copy.py
@@ -312,7 +312,14 @@ class K8SCopyToPod(K8SCopy):
         if self.named_temp_file:
             self.named_temp_file.close()
 
-    def _fail(self, response, msg, **kwargs):
+    def _fail(self, response, msg, stderr=(), **kwargs):
+        """Fail, keeping whatever the remote already wrote to stderr.
+
+        An early close or a stall is usually the symptom, not the cause: tar
+        has typically said why on stderr first.
+        """
+        if stderr:
+            msg = "{0}: {1}".format(msg, "".join(stderr))
         response.close()
         self.close_temp_file()
         self.module.fail_json(
@@ -392,32 +399,26 @@ class K8SCopyToPod(K8SCopy):
             if response.peek_stderr():
                 stderr.append(response.read_stderr().rstrip("\n"))
 
-        def fail(msg):
-            """Fail, keeping whatever the remote already wrote to stderr.
-
-            An early close or a stall is usually the symptom, not the cause:
-            tar has typically said why on stderr first.
-            """
-            if stderr:
-                msg = "{0}: {1}".format(msg, "".join(stderr))
-            self._fail(response, msg)
-
         deadline = time.monotonic() + self.copy_timeout
 
         chunk = tar_buffer.read(self.CHUNK_SIZE)
         while chunk:
             if not response.is_open():
-                fail(
+                self._fail(
+                    response,
                     "connection to Pod {0}/{1} closed before the whole archive was"
                     " sent, the remote file is incomplete".format(
                         self.namespace, self.name
-                    )
+                    ),
+                    stderr,
                 )
             if time.monotonic() > deadline:
-                fail(
+                self._fail(
+                    response,
                     "timed out after {0}s sending the archive to Pod {1}/{2}".format(
                         self.copy_timeout, self.namespace, self.name
-                    )
+                    ),
+                    stderr,
                 )
             # Non-blocking, keeps the receive buffer clear while we write.
             drain(0)
@@ -435,9 +436,11 @@ class K8SCopyToPod(K8SCopy):
         # exit, and have its status reported on the error channel.
         while response.is_open():
             if time.monotonic() > deadline:
-                fail(
+                self._fail(
+                    response,
                     "timed out after {0}s waiting for tar to finish extracting on Pod"
-                    " {1}/{2}".format(self.copy_timeout, self.namespace, self.name)
+                    " {1}/{2}".format(self.copy_timeout, self.namespace, self.name),
+                    stderr,
                 )
             drain(1)
 
--- a/plugins/module_utils/copy.py
+++ b/plugins/module_utils/copy.py
@@ -392,24 +392,32 @@ class K8SCopyToPod(K8SCopy):
             if response.peek_stderr():
                 stderr.append(response.read_stderr().rstrip("\n"))
 
+        def fail(msg):
+            """Fail, keeping whatever the remote already wrote to stderr.
+
+            An early close or a stall is usually the symptom, not the cause:
+            tar has typically said why on stderr first.
+            """
+            if stderr:
+                msg = "{0}: {1}".format(msg, "".join(stderr))
+            self._fail(response, msg)
+
         deadline = time.monotonic() + self.copy_timeout
 
         chunk = tar_buffer.read(self.CHUNK_SIZE)
         while chunk:
             if not response.is_open():
-                self._fail(
-                    response,
+                fail(
                     "connection to Pod {0}/{1} closed before the whole archive was"
                     " sent, the remote file is incomplete".format(
                         self.namespace, self.name
-                    ),
+                    )
                 )
             if time.monotonic() > deadline:
-                self._fail(
-                    response,
+                fail(
                     "timed out after {0}s sending the archive to Pod {1}/{2}".format(
                         self.copy_timeout, self.namespace, self.name
-                    ),
+                    )
                 )
             # Non-blocking, keeps the receive buffer clear while we write.
             drain(0)
@@ -427,10 +435,9 @@ class K8SCopyToPod(K8SCopy):
         # exit, and have its status reported on the error channel.
         while response.is_open():
             if time.monotonic() > deadline:
-                self._fail(
-                    response,
+                fail(
                     "timed out after {0}s waiting for tar to finish extracting on Pod"
-                    " {1}/{2}".format(self.copy_timeout, self.namespace, self.name),
+                    " {1}/{2}".format(self.copy_timeout, self.namespace, self.name)
                 )
             drain(1)
 
--- a/plugins/module_utils/copy.py
+++ b/plugins/module_utils/copy.py
@@ -18,7 +18,9 @@ from __future__ import absolute_import, division, print_function
 __metaclass__ = type
 
 import os
+import shlex
 import tarfile
+import time
 from abc import ABCMeta, abstractmethod
 from select import select
 from tempfile import NamedTemporaryFile, TemporaryFile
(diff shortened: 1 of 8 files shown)


tests/unit/module_utils/test_copy.py
```diff
@@ -0,0 +1,272 @@
...
+    """The connection must not be closed until the remote process is done."""
+    copier = make_copier()
+    response = FakeWSClient(close_after=5)
+
+    error, stdout, stderr = stream_archive(
+        copier, response, 3 * K8SCopyToPod.CHUNK_SIZE
+    )
+
+    assert error["status"] == "Success"
+    assert stdout == [] and stderr == []
```
beeankha:
```suggestion
    assert stdout == []
    assert stderr == []
```

Seems like SonarCloud prefers the assertions get listed separately like this.

yurnov: Changed as a separate commit

plugins/module_utils/copy.py
```diff
@@ -296,14 +298,146 @@ class K8SCopyToPod(K8SCopy):
...
+        else:
+            command = ["tar", "-xmf", "-"]
+
+        if dest_file.startswith("/"):
+            command.extend(["-C", "/"])
+        return command
+
+    def can_bound_stream(self):
+        """Whether the container can run ``sh -c 'head -c ...'``."""
+        error, _out, _err = self._run_from_pod(
```
oraNod, line 348: @yurnov I've conducted a review of these changes with Claude because I'm still pretty new to this codebase. I'll share the findings and hope that they are helpful but please take them with a pinch of salt.

The pre-flight `can_bound_stream()` probe still calls the old, un-timeout'd `_run_from_pod()`, so it can hang forever even though the whole point of this PR is to bound hangs with `copy_timeout`.

plugins/module_utils/copy.py
```diff
@@ -296,14 +298,146 @@ class K8SCopyToPod(K8SCopy):
...
+            if time.monotonic() > deadline:
+                self._fail(
+                    response,
+                    "timed out after {0}s sending the archive to Pod {1}/{2}".format(
+                        self.copy_timeout, self.namespace, self.name
+                    ),
+                )
+            # Non-blocking, keeps the receive buffer clear while we write.
+            drain(0)
+            response.write_stdin(chunk)
```
oraNod, line 425: `write_stdin()` isn't exception-guarded; a socket drop right after the `is_open()` check bypasses the `specific _fail()` cleanup/message and leaks the unclosed exec response.

plugins/module_utils/copy.py
```diff
@@ -296,14 +298,146 @@ class K8SCopyToPod(K8SCopy):
...
+            if response.peek_stdout():
+                stdout.append(response.read_stdout().rstrip("\n"))
+            if response.peek_stderr():
+                stderr.append(response.read_stderr().rstrip("\n"))
+
+        deadline = time.monotonic() + self.copy_timeout
+
+        chunk = tar_buffer.read(self.CHUNK_SIZE)
+        while chunk:
+            if not response.is_open():
```
oraNod, line 406: stderr already captured by `drain()` is discarded when `_fail()` reports an early-close/timeout, so the real remote error (e.g. "read-only filesystem") is replaced by a generic message.

oraNod (approved):
Aside from the comments that I've added from the Claude review everything looks good to me and I can follow what's going on with the changes. Thank you @yurnov

plugins/module_utils/copy.py
yurnov: True, `_run_from_pod()` has no deadline. It never had one though, and `run()` already calls it twice before the probe on every `to_pod` copy, through `is_directory_path_from_pod()` (`test -e`, `test -d`), so the probe adds a third such call rather than a new way to hang. The probe itself is `sh -c 'command -v head'`, which returns immediately.

`copy_timeout` is scoped to the streaming and the remote `tar` because that is where #776 hangs. Bounding `_run_from_pod()` means putting a deadline in the shared base class, which the `from_pod` path uses too -- I'd rather not widen this PR into that. Happy to open a follow-up issue for it if you want.

plugins/module_utils/copy.py
yurnov: The TOCTOU is real, but the consequences are milder than that. `k8s_cp.execute_module()` already wraps `k8s_copy.run()` in `except Exception` and turns it into `fail_json`, so a socket drop in `write_stdin()` fails the task with the websocket error rather than a traceback, and the interpreter exiting closes the socket and the temp file. What's lost is the more specific message, not the cleanup. `write_stdin()` was unguarded on `main` before this PR too, so no change in behaviour here.

plugins/module_utils/copy.py
yurnov: Good catch -- fixed. The three in-stream failures now go through a local `fail()` that appends whatever `drain()` collected, so a read-only mount or a missing directory shows up instead of just "the remote file is incomplete". Added `test_early_close_reports_what_tar_said` for it.

yurnov:
> Aside from the comments that I've added from the Claude review everything looks good to me and I can follow what's going on with the changes. Thank you @yurnov

Answered to each comment (and fixed one finding)

yurnov:
Pushed refactoring to avoid complains from QonarQube

plugins/module_utils/copy.py
oraNod: Yeah let's not widen the scope of the changes here. Honestly I'll leave it up to you to open a follow on if you feel that is worthwhile since you know the codebase much better. Thanks!

oraNod:
recheck

tests/integration/targets/k8s_copy/tasks/test_copy_repeated.yml
```diff
@@ -0,0 +1,45 @@
...
+  block:
+  - set_fact:
+      test_directory: /tmp/test_k8scp_repeated
+      copy_iterations: 10
+    no_log: true
+
+  - name: create temporary directory for local files
+    ansible.builtin.file:
+      path: "{{ test_directory }}"
+      state: directory
```
abikouo:
```suggestion
  - name: Create temporary directory
     ansible.builtin.tempfile:
       state: directory
       suffix: k8scp_repeated
     register: test_dir

  - name: Set variable for test directory
     ansible.builtin.file:
       test_directory: "{{ test_dir.path }}"
```

tests/integration/targets/k8s_copy/tasks/verify_one_copy.yml
```diff
@@ -0,0 +1,22 @@
+---
+# One iteration of the repeated-copy regression check, see test_copy_repeated.yml
+- name: remove any file left by the previous iteration
+  k8s_exec:
+    namespace: '{{ copy_namespace }}'
+    pod: '{{ pod_with_one_container.name }}'
+    command: 'rm -f /repeated.bin'
```
abikouo, line 7:
```suggestion
    command: 'rm -f /repeated.bin'
 ignore_errors: true
```

abikouo (approved):

tests/integration/targets/k8s_copy/tasks/verify_one_copy.yml
yurnov: Done

tests/integration/targets/k8s_copy/tasks/test_copy_repeated.yml
yurnov: Done

oraNod:
Thanks @yurnov

Issue #1146, opened 2026-06-12
Helm 4.2.1 moves registry success messages (Login Succeeded / Pushed / Pulled / Digest) from `stderr` to `stdout` — update docs and tests


yurnov:
##### SUMMARY

Helm currently writes non-error success messages for OCI registry commands (`helm registry login`, `helm push`, `helm pull`) to **stderr** instead of stdout — see helm/helm#13464. The kubernetes.core helm modules, their `RETURN` documentation, and the integration tests were written around that quirk.

Helm is fixing this in helm/helm#32056 (milestone **4.2.1**): the messages `Login Succeeded`, `Pushed:`, `Pulled:`, and `Digest:` will move from **stderr to stdout**. Once CI runners pick up Helm >= 4.2.1, some of our docs/tests become inaccurate or fail.

This is forward-compatibility hardening — there is **no functional bug** in the modules themselves (see analysis below).

##### COMPONENT NAME

- `helm_registry_auth`
- `helm_pull`
- docs + integration tests

##### ANALYSIS / IMPACT

**Module logic — no impact.** Both modules key on the return code, not on parsing success-message text:
- `helm_registry_auth` derives `changed` from `state` and success from `rc`; the only stderr string it inspects is the *error* `"Error: not logged in"`, which stays on stderr.
- `helm_pull` keys on `rc == 0` and does filesystem-based idempotency; it returns stdout/stderr verbatim.
- There is no `helm_push` module; `helm push` is only exercised via `shell` in the `helm_registry_auth` integration test.

**Documentation — needs update.**
- `plugins/modules/helm_registry_auth.py` `RETURN`: `stderr` carries `sample: 'Login Succeeded\n'` plus a note that helm prints success to stderr. This becomes version-dependent/misleading on Helm >= 4.2.1 (the message will be on stdout).

**Tests — one real breakage, one to verify** (in `tests/integration/targets/helm_registry_auth/tasks/run_tests.yml`):
- `'Login Succeeded' in ...stdout_lines + ...stderr_lines` — already robust (checks the union).
- `'Pushed: ...' in _save_chart.stderr` — **breaks** on Helm >= 4.2.1 (moves to stdout). 
- `'Removing login credentials' in ..._logout.stderr` — **to verify**: logout shares the registry client the PR rewires, so its message may move to stdout as well.
- Error-message assertions (`push access denied`, `basic credential not found`, `401`, `unauthorized`) are unaffected — errors stay on stderr. 

[...]

PR #1223, opened 2026-08-17
fix(helm): skip Helm's OCI registry progress messages on stdout

### LINKED ISSUE #1222: kubernetes.core.helm with an OCI registry and a version with a + sign will fail with Helm >= 4.2.1
Hi,

When using `kubernetes.core.helm 6.5.0` on an OCI registry and a chart with a version with a `+` sign like `109.0.1+up1.0.2` will fail when using Helm >= 4.2.1 (4.2.0 does work) with this error:
```
TASK [helm : Install "xxxxx" helm chart] ********************************************************************************************
[ERROR]: Task failed: Module failed: while scanning a simple key
  in "<unicode string>", line 3, column 1:
    registry.xxxx.com/yyyyyyy/charts ...
    ^
could not find expected ':'
  in "<unicode string>", line 5, column 1:
    OCI artifact references (e.g. ta ...
    ^
Origin: /Users/xxxxxxxx/_common/_ansible/_yaml/roles/helm/tasks/installChart.yaml:65:3

63 # Regarding update_repo_cache, if the cache is empty,
64 # Helm will fail will an error and stops
65 - name: Install "{{ installName }}" helm chart
     ^ column 3

fatal: [demo19-m1]: FAILED! => {"attempts": 5, "changed": false, "msg": "Task failed: Module failed: while scanning a simple key\n  in \"<unicode string>\", line 3, column 1:\n    registry.xxxxx.com/yyyyyy/charts ... \n    ^\ncould not find expected ':'\n  in \"<unicode string>\", line 5, column 1:\n    OCI artifact references (e.g. ta ... \n    ^"}
``` 

I found some articles mentioning to replace the `+` sign with an underscore, but then an `Semantic Versioning constraints` error will be shown. 

When installing the same chart from the command line with Helm 4.2.4 will work fine, so it looks like an Ansible issue

yurnov:
##### SUMMARY

Helm 4.2.1 moved the registry client's progress messages from stderr to stdout ([helm/helm#32056](https://github.com/helm/helm/pull/32056)). Two of our code paths read that stdout, so on an OCI reference they now get the messages ahead of the real output:

- `helm` — `fetch_chart_info()` parses `helm show chart` output as YAML
- `helm_template` — returns the stdout of `helm template` as its documented result

Neither can notice this from the return code: both commands exit `0`.

Helm rewrites `+` to `_` in OCI tags, and when the reference contains an underscore it also prints a four-line free-text explanation. That is not valid YAML, so installing a chart whose version carries semver build metadata failed outright with `could not find expected ':'`. Without the `+`, `Pulled:` and `Digest:` parse as a mapping instead, so the chart info silently gained `Pulled`/`Digest` keys on every OCI install.

This is the other half of #1146. PR #1147 fixed `helm_registry_auth` on the reasoning that "the modules key on the return code, so logic is unaffected" — true there, but not for these two paths, and neither file was touched.

Two design decisions worth calling out:

**Strip the known messages rather than seek the start of the real output.** My first attempt anchored on the first `chart.Metadata` key. Two findings killed it. `apiVersion` is not reliably first — Helm marshals Chart.yaml with sorted keys, so a chart carrying `annotations:` puts that line ahead of it. And the real output can be *empty*: `helm template --show-only` with no match returns only the messages, so a payload anchor finds nothing and hands them back as though they were rendered templates. Against three payload shapes (`Chart.yaml` keys, `---`, `wrote …`) plus empty, the messages are the stable thing — always at the front, always byte-identical.

**Match Helm's explanation verbatim.** The four-line paragraph is compared against a copy of Helm's own `registryUnderscoreMessage` const rather than matched by pattern or line count. If Helm rewords it, the comparison fails and we are back to today's loud error instead of silently consuming a line of real output. There is a unit test pinning that.

`helm_pull` is deliberately left alone: its `stdout` is documented as "Full `helm pull` command stdout … examine the event log", so the messages legitimately belong there.

Fixes #1222

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

helm
helm_template

##### ADDITIONAL INFORMATION

The reported failure, reproduced on kind against a local `registry:2` with a chart versioned `109.0.1+up1.0.2`:

```
File ".../plugins/modules/helm.py", line 1135, in main
File ".../plugins/modules/helm.py", line 606, in fetch_chart_info
yaml.scanner.ScannerError: while scanning a simple key
  in "<unicode string>", line 3, column 1:
    localhost:5000/charts/demochart: ...
    ^
could not find expected ':'
  in "<unicode string>", line 5, column 1:
    OCI artifact references (e.g. ta ...
    ^
```

Same command on 4.2.0 vs 4.2.2, stdout only — on 4.2.0 the first eight lines went to stderr:

```
 1  Pulled: localhost:5000/charts/demochart:109.0.1_up1.0.2
 2  Digest: sha256:41e9b37f85fa7f5df5fc7f2b7b068eb5...
 3  localhost:5000/charts/demochart:109.0.1_up1.0.2 contains an underscore.
 4
 5  OCI artifact references (e.g. tags) do not support the plus sign (+). To support
 6  storing semantic versions, Helm adopts the convention of changing plus (+) to
 7  an underscore (_) in chart version tags when pushing to a registry and back to
 8  a plus (+) when pulling from a registry.
 9  apiVersion: v2
10  ...
```

Verified across Helm 4.2.0 / 4.2.1 / 4.2.2 on kind, over both a plain-HTTP and a TLS OCI registry, with and without `+`, for a chart carrying `annotations`, and for a local chart directory — checking install, idempotency, and that `helm_template` output parses via `from_yaml_all`.

`helmdiff_check()` is **not** affected, which I expected it to be: `helm diff` is a separate binary doing its own chart loading, and `diff upgrade` of an unchanged OCI release gives 0 bytes on stdout on both 4.2.0 and 4.2.2 (tested with helm-diff 3.15.11). Idempotency via helm-diff never broke.

##### TESTS

The second commit adds `tests/integration/targets/helm_oci_registry_progress`. It pushes a chart at `1.0.0+build.1` and at `1.0.1`, then asserts that `helm` installs it with `status.chart` reporting the version verbatim and a second run is `changed=false`, and that `helm_template` stdout carries no `Pulled:`/`Digest:`/underscore text, starts with `---`, and parses to exactly `['ConfigMap']`. It also asserts Helm really did rewrite `+` to `_` in the pushed tag, so the target fails loudly if it ever stops covering the case it was written for.

**The target is pinned to Helm v4.2.4**, because it is only meaningful on >= 4.2.1. That is also why this went unnoticed: `helm_v*` tops out at v4.0.0, and `install_helm` defaults to v3.16.4, which is what `helm_diff` — until now our only target touching an OCI registry — runs. I will open a separate issue about testing against current Helm, since that gap is what let both #1146 and #1222 reach users.

Confirmed the integration target **fails on the first commit's parent** and passes on the fix, so it does catch the regression.

Unit tests: `tests/unit/module_utils/helm/test_strip_registry_progress.py`, 20 cases — the strip matrix (4 message combinations × 4 payload shapes including empty), `annotations` before `apiVersion`, the silent-pollution case, a reworded paragraph being left alone, and a payload line that merely looks like a registry message. Full unit suite 278 passed; `black`, `isort`, `flake8` and `yamllint -s .` clean.

If the splitter does not pick the new target up from the changed files, `test-all-the-targets` would force it.

[AI-assisted PR]

### DIFF
--- a/changelogs/fragments/20260817-helm-oci-registry-progress-on-stdout.yml
+++ b/changelogs/fragments/20260817-helm-oci-registry-progress-on-stdout.yml
@@ -1,0 +1,3 @@
+bugfixes:
+  - helm - skip Helm's OCI registry progress messages when parsing the output of ``helm show chart``. As of Helm 4.2.1 these are printed to stdout instead of stderr (https://github.com/helm/helm/pull/32056), which made installing a chart from an OCI registry fail with a YAML scanner error when the chart version contained a ``+`` (https://github.com/ansible-collections/kubernetes.core/pull/1223).
+  - helm_template - strip Helm's OCI registry progress messages from the returned ``stdout``, which is documented to contain only the rendered templates (https://github.com/ansible-collections/kubernetes.core/pull/1223).
--- a/plugins/module_utils/helm.py
+++ b/plugins/module_utils/helm.py
@@ -32,6 +32,53 @@ except ImportError:
     HAS_YAML = False
 
 
+# Helm's registry client writes these progress messages to stdout as of helm 4.2.1
+# (helm/helm#32056); earlier versions sent them to stderr. Anything that parses the
+# stdout of a helm command which may pull from an OCI registry has to skip them.
+# Kept verbatim from helm's pkg/registry/client.go so that a reworded message fails
+# to match, and therefore fails loudly, rather than silently eating real output.
+REGISTRY_PROGRESS_RE = re.compile(r"^(?:Pulled|Pushed|Digest): \S")
+
+REGISTRY_UNDERSCORE_SUFFIX = " contains an underscore.\n"
+
+REGISTRY_UNDERSCORE_MESSAGE = """
+OCI artifact references (e.g. tags) do not support the plus sign (+). To support
+storing semantic versions, Helm adopts the convention of changing plus (+) to
+an underscore (_) in chart version tags when pushing to a registry and back to
+a plus (+) when pulling from a registry.
+"""
+
+REGISTRY_UNDERSCORE_MESSAGE_LINES = len(REGISTRY_UNDERSCORE_MESSAGE.splitlines())
+
+
+def strip_registry_progress(out):
+    """
+    Drop OCI registry progress messages from the front of a helm command's stdout.
+
+    Helm emits 'Pulled:'/'Pushed:' and 'Digest:' lines, followed - when the
+    reference contains an underscore - by a free-text explanation that is not
+    valid YAML. Since helm 4.2.1 these land on stdout ahead of the real output,
+    so 'helm show chart' can no longer be parsed and 'helm template' no longer
+    returns only rendered manifests.
+    """
+    lines = out.splitlines(keepends=True)
+    index = 0
+    while index < len(lines):
+        if REGISTRY_PROGRESS_RE.match(lines[index]):
+            index += 1
+        elif lines[index].endswith(REGISTRY_UNDERSCORE_SUFFIX):
+            # helm prints its registryUnderscoreMessage right after this line, led
+            # by a blank line. Anything else is not a message we know about.
+            start = index + 1
+            end = start + REGISTRY_UNDERSCORE_MESSAGE_LINES
+            if "".join(lines[start:end]) != REGISTRY_UNDERSCORE_MESSAGE:
+                break
+            index = end
+        else:
+            break
+    return "".join(lines[index:])
+
+
 def parse_helm_plugin_list(output=None):
     """
     Parse `helm plugin list`, return list of plugins
--- a/plugins/modules/helm.py
+++ b/plugins/modules/helm.py
@@ -513,6 +513,7 @@ from ansible.module_utils.basic import missing_required_lib
 from ansible_collections.kubernetes.core.plugins.module_utils.helm import (
     AnsibleHelmModule,
     parse_helm_plugin_list,
+    strip_registry_progress,
 )
 from ansible_collections.kubernetes.core.plugins.module_utils.helm_args_common import (
     HELM_AUTH_ARG_SPEC,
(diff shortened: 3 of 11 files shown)


yurnov:
Hi @beeankha, @abikouo, could I ask you to add `test-all-the-targets` tag and review this PR? And `backport-6`, it should not be backported to `stable-5,` as version 5 does not support Helm v4

yurnov:
Test failed on the infra intermittent issue (timeout accessing chart repo); the test just needs to be re-run

yurnov:
CI failed due to [GitHub Incident](https://www.githubstatus.com/incidents/zkxwbgr0cnmx), it worth to re-run

yurnov:
recheck

beeankha (approved):

oraNod (approved):
This one looks pretty solid to me @yurnov 

I'm not 100% sure if it's related but I believe there's another raw helm stdout here: https://github.com/ansible-collections/kubernetes.core/blob/d9e0218dbae23a183aef829847dfeccf0e4e7535/plugins/modules/helm_pull.py#L512

Might be worth giving that the same treatment? I just went through the helm modules and looked for anything that seemed to return stdout.

yurnov:
> This one looks pretty solid to me @yurnov
> 
> I'm not 100% sure if it's related but I believe there's another raw helm stdout here:
> 
> https://github.com/ansible-collections/kubernetes.core/blob/d9e0218dbae23a183aef829847dfeccf0e4e7535/plugins/modules/helm_pull.py#L512
> 
> Might be worth giving that the same treatment? I just went through the helm modules and looked for anything that seemed to return stdout.

I think that here is as it should be, module return helm output that is documented behavior.

oraNod:
Thanks very much for the fix @yurnov

PR #1225, opened 2026-08-21
tests: update sanity ignores to skip imports

oraNod:
##### SUMMARY

During downstream certification checks, `ansible-test sanity` fails on Python 3.8, 3.9, and 3.10. See errors below. This change updates the sanity ignore files for 2.16 to 2.20 to skip the import for the affected modules.

```
Running sanity test "import" on Python 3.8
Run command with data: importer.py
ERROR: Found 3 import issue(s) on python 3.8 which need to be resolved:
ERROR: plugins/module_utils/client/discovery.py:23:0: traceback: ModuleNotFoundError: No module named 'kubernetes'
ERROR: plugins/module_utils/client/resource.py:21:0: traceback: ModuleNotFoundError: No module named 'kubernetes'
ERROR: plugins/module_utils/k8sdynamicclient.py:25:0: traceback: ModuleNotFoundError: No module named 'kubernetes'

Running sanity test "import" on Python 3.9
Run command with data: importer.py
ERROR: Found 3 import issue(s) on python 3.9 which need to be resolved:
ERROR: plugins/module_utils/client/discovery.py:23:0: traceback: ModuleNotFoundError: No module named 'kubernetes'
ERROR: plugins/module_utils/client/resource.py:21:0: traceback: ModuleNotFoundError: No module named 'kubernetes'
ERROR: plugins/module_utils/k8sdynamicclient.py:25:0: traceback: ModuleNotFoundError: No module named 'kubernetes'

Running sanity test "import" on Python 3.10
Run command with data: importer.py
ERROR: Found 3 import issue(s) on python 3.10 which need to be resolved:
ERROR: plugins/module_utils/client/discovery.py:23:0: traceback: ModuleNotFoundError: No module named 'kubernetes'
ERROR: plugins/module_utils/client/resource.py:21:0: traceback: ModuleNotFoundError: No module named 'kubernetes'
ERROR: plugins/module_utils/k8sdynamicclient.py:25:0: traceback: ModuleNotFoundError: No module named 'kubernetes'
```

##### ISSUE TYPE

- Bugfix Pull Request


yurnov:
I'm not sure that worth to care about error on Python version lower than 3.10 and it worth to exclude them from testing, as collections require ansible-core >= 2.16, and 2.16 support python 3.10-3.12, later version of ansible-core moved this range to 3.12-3.14.

All "No module named 'kubernetes'" is expected in CI/test env where dependency is not installed and can be safely skipped, and they are already skipped for all supported version of python, I don't see a reason to add it for Python 3.8 or 3.9 as they aren't supported.

It worth to create Issue on the sertification tool to avoid a run test with the unsupported matrix combination rather than add them in the multiple collections

yurnov:
Correcting my earlier comment: the `3.10-3.12` I quoted for 2.16 is the **control node** range. The `import` test on `plugins/module_utils/**` is a target test, and 2.16's target range is `2.7, 3.6-3.12`, so 3.8/3.9 aren't unsupported from the test perspective. However, README.md claim that version of Python lower that 3.9 is not supported since version 3.0.0.

I dug into this properly and opened #1226 as an alternative.

Root cause is `tests/config.yml`: it declares `python_requires: ">=3.6"` while the README says "Collection supports 3.9+" on `main`, `stable-6` and `stable-5` (and also on `stable-3` and `stable-4`, this was added to version 3.0.0, but only mentioned 3 branches are important now). That declaration is what pulls 3.6/3.7/3.8 into the matrix. #1226 sets it to `">=3.9"` and adds the entries for the target versions that genuinely remain in range.

Three reasons I think that's the better shape:

1. This patch doesn't make `stable-2.16` green. I ran it: with `>=3.6` the import test there runs on 3.6-3.12 and the 8.12.0 container ships 3.6/3.7, so adding only `import-3.8!skip` leaves both failing. (The `3.8, 3.9, 3.10` log in the description reproduces exactly on `stable-2.18`)
2. `ignore-2.21.txt` / `ignore-2.22.txt` aren't updated here, so they break as soon as those branches become the pinned EE targets.
3. Leaving `>=3.6` in place keeps claiming the modules support Python 3.6, and the ignore files have to be re-extended each time a branch shifts its target range.

I'm open to discussion.

oraNod:
@yurnov Thanks so much for digging into this and opening #1226 

That's a good find with the root cause being in `tests/config.yml`. I definitely agree that updating the Python version there is the right move.

I also acknowledge your point that my PR does not update the `ignore-2.21.txt` and `ignore-2.22.txt` files. Those would break as soon as the EE targets get updated, as you say. My plan was to handle those in a separate PR as a companion to this one but I should have left a comment to make that clear. I wanted to dig into the Python versions for those files a bit more.

Anyway I'm happy to close this now in favour of your changes. Once again, thank you for finding the right shape for the changes.

PR #1230, opened 2026-08-26
fix(helm): validate reuse_values against reset_values and replace

### LINKED ISSUE #1229: Question about helm_module reuse_values and reset_values on the documentation exemple
##### SUMMARY

In the documentation for reuse_values it mention that it reuse values only if reset_values is false. But reset_values is true by default.

On the example part for reuse_values, there is not the reset_values to false: https://docs.ansible.com/projects/ansible/latest/collections/kubernetes/core/helm_module.html

```yaml
# Deploy latest version
- name: Deploy latest version of Grafana chart using reuse_values
  kubernetes.core.helm:
    name: test
    chart_ref: stable/grafana
    release_namespace: monitoring
    reuse_values: true
    values:
      replicas: 2
      version: 3e8ec0b2dffa40fb97d5342e4af887de95faa8c61a62480dd7f8aa03dffcf533
```

Is it on purpose or can it be added ?

(i looked quickly on the example to use the reuse_values but when i ran the playbook, my helm values was reset)

##### ISSUE TYPE
- Documentation Report

##### COMPONENT NAME

helm_module

##### ANSIBLE VERSION

```paste below
ansible [core 2.21.3]
  config file = /home/me/path/to/project/ansible.cfg
  configured module search path = ['/home/me/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/me/path/to/project/.venv/lib/python3.13/site-packages/ansible
  ansible collection location = /home/me/.ansible/collections:/usr/share/ansible/collections
  executable location = ./.venv/bin/ansible
  python version = 3.13.5 (main, Jul 15 2026, 20:25:40) [GCC 14.2.0] (/home/me/path/to/project/.venv/bin/python3)
  jinja version = 3.1.6
  pyyaml version = 6.0.3 (with libyaml v0.2.5)
```

yurnov:
##### SUMMARY

`reset_values` defaults to `true`, and helm ignores `--reuse-values` whenever `--reset-values` is passed. The `reuse_values` example in documentation set only `reuse_values: true`, so following it reset the release's values (which is what #1229 reports).

Every integration test we have already pairs the two correctly (`test_helm_reuse_values.yml`, `helm_diff/tasks/reuse_values.yml`), and `helm_diff/tasks/reuse_values.yml` even asserts the trap directly: `reuse_values: true` with the default `reset_values` reports `changed`. The behaviour is intentional and covered; only the example contradicted it.

`reset_then_reuse_values` has the same shape and is also a no-op out of the box, since helm ignores it when either `--reset-values` or `--reuse-values` is set. It had no example at all.

Looking at the rest of the parameter group turned up a second, unrelated-looking bug in the same code: `--reuse-values` and `--reset-then-reuse-values` were appended *after* the install/upgrade branch in `deploy()`, so they leaked into the `replace` path. `helm install` accepts neither flag, in v3 or v4, so `replace: true` with either option could only ever produce `Error: unknown flag`. Because the guard was `if reuse_values is not None`, even an explicit `reuse_values: false` broke the task.

Two decisions worth calling out:

**Warn, do not fail.** Every one of these combinations is accepted by playbooks today, so failing on any of them would turn passing tasks red on upgrade. Flipping the `reset_values` default would silently change what existing playbooks deploy, and the `replace` combinations only ever failed once something changed -- a converged release never reaches `deploy()`, so helm never saw the bad flag. Since `deploy()` now keeps all three flags inside the `helm upgrade` branch, dropping them under `replace` leaves a valid `helm install` command behind, so warning costs nothing. `cleanup_on_fail` still fails hard on `replace`, which looks inconsistent: it is `version_added: 6.6.0` and has not shipped, so strictness there cannot regress anyone, while `reuse_values` has been accepted since 3.0.0.

**Validate in `main()`, not in `deploy()`.** `cleanup_on_fail` checks inside `deploy()`, but that path is skipped when the idempotency check reports no change -- so the warning would never reach the one case that most needs it, a release that is already converged and looks fine. Validating on the `release_state: present` path covers `deploy()`, `helmdiff_check()` and check mode alike.

`deploy()` now keeps `--reset-values`, `--reuse-values` and `--reset-then-reuse-values` inside the `helm upgrade` branch, and `helmdiff_check()` guards the latter two with `not replace`, matching the guard already on `--reset-values`. Both are load-bearing: warning rather than rejecting means these guards are what actually drops the flags from the generated commands, and removing either fails `TestValueReuseOptionsWithHelmDiff`.

Two adjacent bugs in the same function are deliberately **not** in this PR, since they concern different parameters and I would rather keep the diff to one thing: `take_ownership` is emitted only in the upgrade branch even though `helm install --take-ownership` exists (v3 >= 3.17 and v4), so `replace: true` silently drops it; and `--force` is appended regardless of branch, which on helm v3 makes `force: true` + `replace: true` another `unknown flag` (v4 is fine, it emits `--force-replace`). Later I will create a separate PR for it.

Fixes #1229

##### ISSUE TYPE

- Bugfix Pull Request
- Documentation Pull Request

##### COMPONENT NAME

helm

##### ADDITIONAL INFORMATION

**No behaviour change to plan for.** `replace: true` combined with `reuse_values` used to succeed whenever the diff reported no change, because `deploy()` was never called, and fail with helm's `unknown flag` as soon as anything did change. It now succeeds in both cases: the flag is dropped and the module warns, naming the option.

Commands the module generates, before and after (helm 3.18.0, fresh install path):

```
########## before
reuse_values: true
  cmd: /usr/bin/helm upgrade -i --reset-values --reuse-values=True test './chart'

reuse_values: true + reset_values: false
  cmd: /usr/bin/helm upgrade -i --reuse-values=True test './chart'

replace: true + reuse_values: true
  cmd: /usr/bin/helm install --reuse-values=True --replace test './chart'
       Error: unknown flag: --reuse-values

########## after
reuse_values: true
  cmd: /usr/bin/helm upgrade -i --reset-values --reuse-values=True test './chart'
  warn: reuse_values is ignored because reset_values is true. Set reset_values=false to make it take effect.

reuse_values: true + reset_values: false
  cmd: /usr/bin/helm upgrade -i --reuse-values=True test './chart'

replace: true + reuse_values: true
  cmd: /usr/bin/helm install --replace test './chart'
  warn: reuse_values is ignored because replace=true deploys through 'helm install', which does not accept the flag.
```

The first case is unchanged on purpose: same command, now with a warning explaining why `--reuse-values` will not do anything.

That `helm install` rejects the flag, on the helm this was checked against:

```
$ helm version --short
v4.2.2+gb05881c
$ helm install --reuse-values=True foo ./nonexistent
Error: unknown flag: --reuse-values
$ helm install --reset-then-reuse-values foo ./nonexistent
Error: unknown flag: --reset-then-reuse-values
```

Neither flag is in the v3 `helm install` options either -- see the [v3 docs](https://v3.helm.sh/docs/helm/helm_install/) -- so this is not a Helm v4 regression, it has been broken since `reuse_values` was added in 3.0.0 (#575).

**Docs.** All three descriptions now state that `reset_values` defaults to `true` and has to be set to `false` for the other two to take effect, and that helm ignores both under `replace`. The `reuse_values` example gained `reset_values: false`, and a `reset_then_reuse_values` example was added. 

**Tests.** `tests/unit/modules/test_module_helm.py`: `TestValueReuseOptions`, 11 cases over the command built on the install path, and `TestValueReuseOptionsWithHelmDiff`, 7 cases over the `helm diff upgrade` command, including the `helm >= 3.14.0` and `helm diff >= 3.9.12` gates that had integration coverage only. There was no unit coverage for any of these three parameters before.

Verified from a clean checkout of the branch: 50/50 helm unit tests pass (ansible-core 2.18.19), `ansible-test sanity --test validate-modules --test yamllint --test pep8 --test pylint` exits 0 on the changed files, and `antsibull-changelog lint` is clean. `black` clean.


plugins/modules/helm.py
```diff
@@ -616,6 +637,52 @@ def fetch_chart_info(
     return yaml.safe_load(out)
 
 
+def validate_value_options(
+    module, replace, reuse_values, reset_values, reset_then_reuse_values
+):
+    """
+    Check the value-reuse options against each other and against 'replace'.
+
+    'helm install' accepts neither '--reuse-values' nor '--reset-then-reuse-values', so
```
oraNod:
```suggestion
    'helm install' does not accept '--reuse-values' or '--reset-then-reuse-values', so
```

oraNod: It's just a nit but this "neither nor" phrasing seems a bit wordy to me.

oraNod (commented):
Just gave this one a quick glance but will come back to it later. Spotted one bit of wordsmithing in the meantime. Cheers.

plugins/modules/helm.py
yurnov: Thanks, English is not my native.

oraNod (approved):
This looks really good to me. Thanks a mill @yurnov

changelogs/fragments/20260826-helm-value-reuse-options.yaml
```diff
@@ -0,0 +1,5 @@
+---
+bugfixes:
+  - helm - do not pass the upgrade-only ``--reuse-values`` and ``--reset-then-reuse-values`` flags to ``helm install``. Combining ``reuse_values`` or ``reset_then_reuse_values`` with ``replace`` produced an invalid command that helm rejected with ``unknown flag``, and is now reported as a module failure instead (https://github.com/ansible-collections/kubernetes.core/pull/1230).
```
abikouo: Could you please split this into 2 separate entries?

plugins/modules/helm.py
```diff
@@ -616,6 +637,52 @@ def fetch_chart_info(
...
+    no-op. Only warn about that, so playbooks relying on the current behaviour keep
+    working.
+    """
+    if replace:
+        for name, requested in (
+            ("reuse_values", reuse_values is not None),
+            ("reset_then_reuse_values", reset_then_reuse_values),
+        ):
+            if requested:
+                module.fail_json(
```
abikouo: Instead of throwing a hard error here, raising a warning might be a better approach. What do you think?

tests/unit/modules/test_module_helm.py
```diff
@@ -890,3 +890,170 @@ def test_cleanup_on_fail_with_replace_fails(self):
                 },
                 "3.17.0",
             )
+
```
abikouo, line 924: We should add test scenarios for the helm-diff function.

abikouo (changes requested):
Good job overall @yurnov , thanks a lot.
Just few remarks

changelogs/fragments/20260826-helm-value-reuse-options.yaml
yurnov: Done, three entries now, two bugfixes and one minor change:

- the `reuse_values` example, which is the bug reported in #1229
- the upgrade-only flags no longer reaching `helm install`
- the warnings, and the documentation that now spells out the `reset_values` dependency

plugins/modules/helm.py
yurnov: Agreed, changed to a warning.

You are right, and my PR description overstated the case for failing. I wrote that any playbook hitting this combination "is already broken", but that is not true from the operator's side: with `replace: true` on an already converged release, `helmdiff_check` reports no change, `deploy()` is never called, helm never sees the flag, and the task is green on every run.

Two things I would rather flag than bury:

`cleanup_on_fail` still fails hard on `replace`, so the module now has two policies for the same class of problem. My reasoning for leaving it: `cleanup_on_fail` is `version_added: 6.6.0` and has not shipped, so strictness there cannot regress anyone, while `reuse_values` has been in place since 3.0.0. If you would rather have one policy I am happy to align them in a separate PR, in either direction.

`reuse_values: true` with `replace` is now a silent fresh install rather than an error, and warnings do get lost in CI output. The window is narrow, since `helm install --replace` refuses a release name that is still in use, but it is not zero. Say the word if you would prefer the failure kept for the `true` case and the warning only for `false`

tests/unit/modules/test_module_helm.py
yurnov: Added `TestValueReuseOptionsWithHelmDiff`, and a `diff_command_for_version` helper next to the existing `deploy_command_for_version`. You were right that there was nothing: `deploy_command_for_version` mocks `get_release_status` to `None`, so every test in the PR went down the fresh-install path and `helmdiff_check` was never entered.

yurnov:
> Merge Failed.
> 
> This change or one of its cross-repo dependencies was unable to be automatically merged with the current state of its repository. Please rebase the change and upload a new patchset.

Going to rebase

abikouo (approved):

oraNod:
@yurnov Looks like there are some integration test failures. Do you mind taking a look please?

oraNod:
I think maybe that integration test was failing for some external reason although I wasn't sure at first glance of the error messages. Let's run it again and see.

oraNod:
Thanks @yurnov

yurnov:
Hi @oraNod, sorry, I was busy with my work task, it seems that the recent API rate limit affects the stability of the integration test

PR #1231, opened 2026-08-31
ci: add the PE certification checker workflow

oraNod:
##### SUMMARY

This change adds the [Red Hat partner certification checker](https://github.com/ansible-collections/partner-certification-checker#red-hat-partner-certification-checker) workflow that runs checks that match the Automation Hub import process.

Tracked in Jira via `ACA-5380`.

##### ISSUE TYPE

- Feature Pull Request

### DIFF
--- a/.github/workflows/certification.yml
+++ b/.github/workflows/certification.yml
@@ -6,7 +6,9 @@ permissions:
 
 on:
   pull_request:
-    branches: [main]
+    branches:
+      - main
+      - stable-*
   workflow_dispatch:
 
 concurrency:
--- a/.github/workflows/certification.yml
+++ b/.github/workflows/certification.yml
@@ -1,0 +1,18 @@
+---
+name: Run collection certification checks
+
+permissions:
+  contents: read
+
+on:
+  pull_request:
+    branches: [main]
+  workflow_dispatch:
+
+concurrency:
+  group: cert-ver-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
+jobs:
+  call:
+    uses: ansible-collections/partner-certification-checker/.github/workflows/certification-reusable.yml@f4bf7ba4d3f008b2db8fc6a812fa3f89fdf42c2e # v4.1.0


.github/workflows/certification.yml
```diff
@@ -0,0 +1,18 @@
...
+    branches: [main]
+  workflow_dispatch:
+
+concurrency:
+  group: cert-ver-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
+jobs:
+  call:
+    uses: ansible-collections/partner-certification-checker/.github/workflows/certification-reusable.yml@f4bf7ba4d3f008b2db8fc6a812fa3f89fdf42c2e # v4.1.0
```
yurnov, line 18: If you add the exact SHA of the version, then it's worth adding Dependabot config to have it updated, like:
```
version: 2
updates:
  - package-ecosystem: "github-actions"
    directory: "/"
    schedule:
      interval: "weekly"
```

Alternatively, you can just use version, like other GitHub Actions in this repo.

oraNod: Thanks for pointing that out. Agree that it would be good to have Dependabot bumping action versions. I've added that to one or two other collection repos already and can handle that as a follow on to this PR.

abikouo (approved):

GomathiselviS (approved):

.github/workflows/certification.yml
```diff
@@ -0,0 +1,18 @@
+---
+name: Run collection certification checks
+
+permissions:
+  contents: read
+
+on:
+  pull_request:
+    branches: [main]
```
oraNod: Should this run on other stable branches? If so I think we'll need to specify which branches instead of using a wildcard like `stable-*`.

yurnov: for me, it better to have `stable-*` rather than `[stable-6, stable-5]`, so that when, somewhere in 2027 will be created `stable-7` (with an announced breaking change) and `stable-5` is marked as EOS/EOL, it will work automatically without any updates here. 

Only maintainers can create branches, so, not a real drawback

yurnov:
One comment: the [Red Hat partner certification checker](https://github.com/ansible-collections/partner-certification-checker#red-hat-partner-certification-checker) workflow runs a sanity check with stable-2.16, stable-2.18, and stable-2.20 (all of them with Python 3.12), that duplicate some of the existing testing. 

It is probably worth considering the removal of stable-2.16, stable-2.18, and stable-2.20 in combination with 3.12 to avoid this duplication and decrease CI cost. On the other hand, this would require maintaining an exclusion list aligned with two upstream repos (ansible-network/github_actions and ansible-collections/partner-certification-checker), and sanity testing is cheap, so probably it worth to keep it as is.

.github/workflows/certification.yml
oraNod: Agree that `stable-*` is better for the reasons you mention. I was thinking about some of the older branches in this repo like the stable-2 ones. I guess, though, it isn't really a problem as those branches are not likely to receive any changes.

oraNod:
> One comment: the [Red Hat partner certification checker](https://github.com/ansible-collections/partner-certification-checker#red-hat-partner-certification-checker) workflow runs a sanity check with stable-2.16, stable-2.18, and stable-2.20 (all of them with Python 3.12), that duplicate some of the existing testing.
> 
> It is probably worth considering the removal of stable-2.16, stable-2.18, and stable-2.20 in combination with 3.12 to avoid this duplication and decrease CI cost. On the other hand, this would require maintaining an exclusion list aligned with two upstream repos (ansible-network/github_actions and ansible-collections/partner-certification-checker), and sanity testing is cheap, so probably it worth to keep it as is.

Thanks for this comment and observation @yurnov While discussing the adoption of this workflow with @GomathiselviS we talked about this duplication. The certification checker is maintained by the Red Hat partner engineering team. Maintaining that exclusion list creates some maintenance overhead and potentially risks creating gaps in test coverage and things getting out of sync.

The certification checker workflow is also intended for collections that are uploaded to Automation Hub as supported while the workflows in github_actions also apply to community only collections. So we'd need to make sure that any exclusion list did not affect community collections.

It isn't very DRY but, like you say, sanity testing is cheap.

.github/workflows/certification.yml
yurnov: Older branches are not maintained anymore, it's mentioned here https://github.com/ansible-collections/kubernetes.core#support (but probably not really clear, as it was a short period of time when ' stable-3 ' and ' stable-5 ' were supported simultaneously)

yurnov (approved):
LGTM

yurnov:
@abikouo @beeankha should this one be backported to stable-5 and stable-6? I think that should, as it's a release branch, and releases should be certified

oraNod:
> @abikouo @beeankha should this one be backported to stable-5 and stable-6? I think that should, as it's a release branch, and releases should be certified

@yurnov I'll go ahead and open the backport PRs and request reviews there since this one is already merged. Thank you!

PR #1232, opened 2026-08-31
CI - Update workflow to run helm test against multiple versions

### LINKED ISSUE #1224: CI does not test against current Helm, so upstream Helm changes reach users before us
##### SUMMARY
Our Helm integration matrix is pinned to versions we chose by hand, and it has fallen behind
upstream. Nothing in CI runs the current Helm release, so a behaviour change in Helm reaches users
as a bug report rather than a red build.

Today's matrix (`tests/integration/targets/helm_v*`):

| target | Helm |
| --- | --- |
| `helm_v3_15_4` | v3.15.4 |
| `helm_v3_16_0` | v3.16.0 |
| `helm_v3_17_0` | v3.17.0 |
| `helm_v4_0_0` | v4.0.0 |

`install_helm` also defaults to `helm_version: v3.16.4`, which is what every target that does not
pin a version gets — including `helm_diff`, our only target that exercises an OCI registry.

Current upstream releases are **v4.2.4** and **v3.21.4**. The newest version we test is v4.0.0.

Two issues from one Helm PR, [helm/helm#32056](https://github.com/helm/helm/pull/32056), which
moved the registry client's messages from stderr to stdout in **Helm 4.2.1**:

- #1146 — `helm_registry_auth` return docs and integration assertions assumed stderr. Fixed in #1147.
- #1222 — `helm` failed outright installing an OCI chart whose version contains a `+`, because `fetch_chart_info()` parses `helm show chart` stdout. `helm_template` was silently returning the same messages inside its documented "rendered templates" output. Both will be fixed in #1223.

Neither was caught by CI, and neither could have been: no target ran a Helm >= 4.2.1, and the only target touching an OCI registry runs v3.16.4.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
tests/integration, CI

##### ADDITIONAL INFORMATION

Roughly in order of value per unit of CI time:

1. Add a target that resolves the latest v3.x and v4.x at runtime rather than pinning, or a `helm_latest` target that install_helm feeds `DESIRED_VERSION` from the GitHub releases API. This is the one that would have caught both issues above.
2. Move `install_helm`'s default off v3.16.4. Every unpinned target silently tests a version from 2024. At minimum bump it; better, make the default the newest v3.
3. Cover OCI registries beyond `helm_diff`. #1222 landed a target that does (`helm_oci_registry_progress`), but OCI is otherwise nearly untested, and it is where Helm has been changing most.

Optionally can be considered a scheduled (weekly or bi-weekly) run against the latest, kept separate from the per-PR matrix, would give the signal without slowing PRs down, and a failure there is information about upstream, not about the PR that happened to be merging.

abikouo:
##### SUMMARY

Closes #1224 

- Update CI to test Helm targets against multiple versions (the latest version is part of the matrix)
- Fix issue on helm module: Remove unsupported parameter ``plain-http`` for ``helm diff upgrade`` command

##### ISSUE TYPE

- Bugfix Pull Request
- Feature Pull Request

##### COMPONENT NAME

- CI
- Helm


yurnov (commented):
Revoked this comment, trying to use AI agent, and seems to have made the wrong observation

.github/workflows/integration-tests-helm.yaml
```diff
@@ -0,0 +1,200 @@
...
+    runs-on: ubuntu-latest
+    outputs:
+      helm_targets: ${{ steps.select-targets.outputs.targets }}
+    steps:
+      - name: Checkout the collection repository
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          path: ${{ env.source_dir }}
+          fetch-depth: "0"
+          ref: ${{ github.event.pull_request.head.sha }}
```
oraNod, line 32:
```suggestion
          ref: ${{ github.event.pull_request.head.sha }}
          persist-credentials: false
```

.github/workflows/integration-tests-helm.yaml
```diff
@@ -0,0 +1,200 @@
...
+            helm-version: "v3.0.0"
+          - ansible_test_target: "helm"
+            helm-version: "v3.1.0"
+    name: "${{ matrix.ansible_test_target }}-${{ matrix.helm-version }}"
+    steps:
+      - name: Checkout kubernetes.core repository
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          path: ${{ env.source }}
+          ref: ${{ github.event.pull_request.head.sha }}
```
oraNod, line 135:
```suggestion
          ref: ${{ github.event.pull_request.head.sha }}
          persist-credentials: false
```

.github/workflows/integration-tests-helm.yaml
```diff
@@ -0,0 +1,200 @@
...
+    types:
+      - opened
+      - reopened
+      - labeled
+      - unlabeled
+      - synchronize
+    branches:
+      - main
+      - stable-*
+
```
oraNod, line 18: Maybe drop permissions at the start and then explicitly set them in jobs where needed? Similar to the changes here: https://github.com/ansible-collections/ansible.mcp/pull/75/changes

.github/workflows/integration-tests.yaml
```diff
@@ -35,18 +34,36 @@ jobs:
...
+        env:
+          ALL_TEST_TARGETS: ${{ steps.splitter.outputs.test_targets }}
+
       - name: Display splitter output
         run: |
-          echo "test_targets=${{ steps.splitter.outputs.test_targets }}"
-          echo "test_targets_json=${{ steps.splitter.outputs.test_targets_json }}"
-          echo "test_jobs=${{ steps.splitter.outputs.test_jobs }}"
+          echo "targets => ${{ steps.select-targets.outputs.targets }}"
+          echo "jobs => ${{ steps.select-targets.outputs.jobs }}"
```
oraNod: This run block could allow code injection via template expansion according to zizmor: https://docs.zizmor.sh/audits/#template-injection

tests/integration/targets/helm/tasks/run_test.yml
```diff
@@ -1,56 +1,53 @@
...
-    - name: test helm upgrade with reuse_values
-      include_tasks: test_helm_reuse_values.yml
+    - name: Test helm upgrade with reuse_values
+      ansible.builtin.include_tasks: test_helm_reuse_values.yml
 
-    - name: test helm upgrade with reset_then_reuse_values
-      include_tasks: test_helm_reset_then_reuse_values.yml
+    - name: Test helm upgrade with reset_then_reuse_values
+      ansible.builtin.include_tasks: test_helm_reset_then_reuse_values.yml
+      when: helm_version | string is ansible.builtin.version('3.14.0')
```
oraNod:
```suggestion
      when: helm_version | string is ansible.builtin.version('3.14.0', '>=')
```

.github/workflows/integration-tests-helm.yaml
```diff
@@ -0,0 +1,200 @@
...
+      - name: Select Helm targets
+        id: select-targets
+        run: |
+          import os, json
+          with open(os.environ.get('GITHUB_OUTPUT'), "a", encoding="utf-8") as fh:
+            data = os.environ.get("ALL_TEST_TARGETS")
+            targets = []
+            for target in data.split(";"):
+              targets.extend([v for v in target.split(':', maxsplit=1)[1].split(',') if v.startswith('helm')])
```
oraNod:
```suggestion
            for target in data.split(";"):
              if target and ':' in target:
                targets.extend([v for v in target.split(':', maxsplit=1)[1].split(',') if v.startswith('helm')])
```

.github/workflows/integration-tests.yaml
```diff
@@ -35,18 +34,36 @@ jobs:
...
+          import os, json
+          with open(os.environ.get('GITHUB_OUTPUT'), "a", encoding="utf-8") as fh:
+            data = os.environ.get("ALL_TEST_TARGETS")
+            index = 0
+            result = {}
+            for target in data.split(";"):
+              k8s = [v for v in target.split(':', maxsplit=1)[1].split(',') if not v.startswith('helm')]
+              if k8s:
+                result[index] = ' '.join(k8s)
+                index += 1
```
oraNod:
```suggestion
            for target in data.split(";"):
              if target and ':' in target:
                k8s = [v for v in target.split(':', maxsplit=1)[1].split(',') if not v.startswith('helm')]
                if k8s:
                  result[index] = ' '.join(k8s)
                  index += 1
```

.github/workflows/integration-tests-helm.yaml
```diff
@@ -0,0 +1,200 @@
...
+          - v3.10.3
+          - v3.15.4
+          - v3.16.0
+          - v3.17.0
+          - v4.0.0
+          - v4.2.1
+          - latest
+        ansible_test_target: ${{ fromJson(needs.splitter.outputs.helm_targets) }}
+        exclude:
+          # helm_registry_auth requires Helm >= 3.8.0
```
oraNod, line 86: Is this comment accurate? I see v3.8.0 and higher in the list of exclusions below.

.github/workflows/integration-tests.yaml
beeankha:
```suggestion
        env:
          TARGETS: ${{ steps.select-targets.outputs.targets }}
          JOBS: ${{ steps.select-targets.outputs.jobs }}
        run: |
          echo "targets => $TARGETS"
          echo "jobs => $JOBS"
 ```

File: .github/workflows/integration-tests-helm.yaml [hunk 9]
```diff
@@ -0,0 +1,200 @@
...
+      - name: Set up Python ${{ env.python-version }}
+        uses: actions/setup-python@5fda3b95a4ea91299a34e894583c3862153e4b97 # v7.0.0
+        with:
+          python-version: ${{ env.python-version }}
+
+      # install ansible
+      - name: Install ansible-core (${{ env.ansible-version }})
+        run: >-
+          python3 -m pip install
+          https://github.com/ansible/ansible/archive/${{ env.ansible-version }}.tar.gz
```
beeankha:
```suggestion
        env:
          ANSIBLE_VERSION: ${{ matrix.ansible-version }}
        run: >-
          python3 -m pip install
          https://github.com/ansible/ansible/archive/$ANSIBLE_VERSION.tar.gz
```

The risk here is pretty low but this change should make the SonarCloud warning go away

.github/workflows/integration-tests.yaml
```diff
@@ -35,18 +34,41 @@ jobs:
...
+                  result[index] = ' '.join(k8s)
+                  index += 1
+            fh.write(f'targets={json.dumps(result)}\n')
+            fh.write(f'jobs={json.dumps(list(result.keys()))}\n')
+        shell: python
+        env:
+          ALL_TEST_TARGETS: ${{ steps.splitter.outputs.test_targets }}
+
       - name: Display splitter output
         run: |
```
oraNod, line 58:
```suggestion
      - name: Display splitter output
```

oraNod: The `run` key got duplicated here.

yurnov: yep, duplicate `run` key breaks integration-tests.yaml entirely

.github/workflows/integration-tests-helm.yaml
```diff
@@ -0,0 +1,205 @@
...
+          persist-credentials: false
+
+      - name: Set up Python ${{ env.python-version }}
+        uses: actions/setup-python@5fda3b95a4ea91299a34e894583c3862153e4b97 # v7.0.0
+        with:
+          python-version: ${{ env.python-version }}
+
+      # install ansible
+      - name: Install ansible-core (${{ env.ansible-version }})
+        env:
```
yurnov, line 145: This workflow's matrix has `helm-version` and `ansible_test_target` only, there is no `ansible-version` key. The version lives in the job `env` (`ansible-version: "milestone"`, line 67), and the step name still reads it from there.

tests/integration/targets/helm/tasks/test_helm_uninstall.yml
```diff
@@ -94,4 +88,4 @@
       name: "{{ helm_namespace }}"
       state: absent
     ignore_errors: true
-  when: test_version.result
+  when: helm_version | string is ansible.builtin.version('3.7.0', '<=')
```
yurnov, line 91: workflow excludes the `helm` target from `v3.0.0` and `v3.1.0`, so the lowest version it ever sees is 3.8.0, so this playbook is never executed

tests/integration/targets/helm_diff/tasks/main.yml
```diff
@@ -266,44 +270,49 @@
...
-      check_mode: true
-      register: _upgrade_2
-
-    - name: Assert that module raised a warning
-      assert:
-        that:
-          - _upgrade_2.changed
-          - _upgrade_2.warnings is not defined
+    - name: Test helm diff with OCI chart
+      when: helm_version is ansible.builtin.version('3.13.0', '<')
```
yurnov, line 286: helm_diff is excluded from v3.0.0, v3.1.0, v3.8.0, v3.10.3 and v4.0.0, so the versions it actually runs on are 3.15.4, 3.16.0, 3.17.0, 4.2.1 and latest, every one of them >= 3.13.0

so, it never executed

.github/workflows/integration-tests-helm.yaml
yurnov: no reason to test it against all versions of helm, but it's worth updating the comment to avoid consufion (i.e. add an additional comment that the version is excluded to make testing cheaper)

plugins/modules/helm.py
yurnov, line 1: It looks unrelated to these PR changes

.github/workflows/integration-tests-helm.yaml
```diff
@@ -0,0 +1,205 @@
...
+    env:
+      source: "./source"
+      ansible_posix: "./ansible_posix"
+      community_general: "./community_general"
+      ansible-version: "milestone"
+      python-version: "3.13"
+    strategy:
+      fail-fast: false
+      matrix:
+        # Ref must match a branch/tag on github.com/ansible/ansible (e.g. stable-2.18, not 2.18).
```
yurnov, line 72: This comment is for Ansible, so it should not be here

yurnov:
Side comment, it's testing/CI-only PR and should not be a blocker for 6.6.0 and 5.4.5

tests/integration/targets/helm_diff/tasks/main.yml
```diff
@@ -257,6 +260,7 @@
         namespace: "{{ helm_namespace }}"
         name: another-chart
         release_values: "{{ test_chart_values }}"
+        plain_http: "{{ helm_version is ansible.builtin.version('3.13.0', '>=') }}"
```
oraNod:
```suggestion
        plain_http: "{{ helm_version | string is ansible.builtin.version('3.13.0', '>=') }}"
```

tests/integration/targets/helm_diff/tasks/main.yml
```diff
@@ -266,44 +270,51 @@
...
```
oraNod:
```suggestion
        plain_http: "{{ helm_version | string is ansible.builtin.version('3.13.0', '>=') }}"
```

tests/integration/targets/helm_diff/tasks/main.yml
```diff
@@ -266,44 +270,51 @@
...
-
-    - name: Assert that module raised a warning
-      assert:
-        that:
-          - _upgrade_2.changed
-          - _upgrade_2.warnings is not defined
+    # helm-diff plugin does not support --plain-http, so this test can only
+    # run against helm versions that do not require it for plain-HTTP registries.
+    - name: Test helm diff with OCI chart
+      when: helm_version is ansible.builtin.version('3.13.0', '<')
```
oraNod:
```suggestion
      when: helm_version | string is ansible.builtin.version('3.13.0', '<')
```

tests/integration/targets/helm_diff/tasks/main.yml
```diff
@@ -266,44 +270,51 @@
...
+      block:
+        - name: Install helm diff
+          helm_plugin:
+            binary_path: "{{ helm_binary }}"
+            state: present
+            plugin_path: https://github.com/databus23/helm-diff
+            plugin_version: "{{ (helm_version is ansible.builtin.version('4.0.0', '>=')) | ternary('3.14.0', '3.4.1') }}"
```
oraNod:
```suggestion
            plugin_version: "{{ (helm_version | string is ansible.builtin.version('4.0.0', '>=')) | ternary('3.14.0', '3.4.1') }}"
```

tests/integration/targets/helm_oci_registry_progress/tasks/main.yml
```diff
@@ -12,15 +12,12 @@
 # messages. Without the fix, the helm task below fails with
 # "could not find expected ':'".
 - name: Test OCI charts whose version contains build metadata
+# Helm >= 4.2.1 is required for this test to be meaningful: that is the release which
+# moved the OCI registry progress messages from stderr to stdout (helm/helm#32056).
+  when: helm_version is ansible.builtin.version('4.2.1', '>=')
```
oraNod:
```suggestion
  when: helm_version | string is ansible.builtin.version('4.2.1', '>=')
```

oraNod (commented):
Thanks for all the work @abikouo Looks great. Only nit is using `| string` filters throughout for version comparison.

PR #1234, opened 2026-09-06
chore(test): add ansible 2.21 to integration tests matrix

yurnov:
##### SUMMARY

The current version of Ansible (at the time of this PR) is 2.21, which is included in the Ansible 14 release.

To ensure testing coverage for the actual version of Ansible, version 2.21 should be included in the integration test matrix.

So, the updated matrix includes:
| version | reson | notes |
| --- | --- | --- |
| stable-2.18 | the latest version supporting turbo mode | Python 3.12 as 3.13 is not supported |
| stable-2.21 | the actual stable version | No turbo mode, only 3.13 to decrease CI cost |
| milestone | the development version |  No turbo mode, only 3.13 to decrease CI cost |
##### ISSUE TYPE

- CI/Testing

##### COMPONENT NAME
.github/workflows/integration-tests.yaml

##### ADDITIONAL INFORMATION
This PR slightly increases the cost and time of CI test execution, and probably worth adding a label `test-all-the-targets` and `skip-changelog`

PR is a result of a conversation with @oraNod

### DIFF
--- a/.github/workflows/integration-tests.yaml
+++ b/.github/workflows/integration-tests.yaml
@@ -61,6 +61,9 @@ jobs:
         enable-turbo-mode: [true, false]
         python-version: ["3.12", "3.13"]
         exclude:
+          # stable-2.18 is used mainly because turbo mode, no reason to test it with turbo mode disabled
+          - ansible-version: "stable-2.18"
+            enable-turbo-mode: false
           # turbo mode is supported only by ansible<2.19
           - ansible-version: "milestone"
             enable-turbo-mode: true
--- a/.github/workflows/integration-tests.yaml
+++ b/.github/workflows/integration-tests.yaml
@@ -56,15 +56,21 @@ jobs:
       fail-fast: false
       matrix:
         # Ref must match a branch/tag on github.com/ansible/ansible (e.g. stable-2.18, not 2.18).
-        ansible-version: ["stable-2.18", "milestone"]
+        #
+        ansible-version: ["stable-2.18", "stable-2.21", "milestone"]
         enable-turbo-mode: [true, false]
         python-version: ["3.12", "3.13"]
         exclude:
+          # turbo mode is supported only by ansible<2.19
           - ansible-version: "milestone"
             enable-turbo-mode: true
-          # stable-2.18 supports 3.12; milestone (devel) requires >=3.13
+          - ansible-version: "stable-2.21"
+            enable-turbo-mode: true
+          # stable-2.18 supports 3.12
           - ansible-version: "stable-2.18"
             python-version: "3.13"
+          - ansible-version: "stable-2.21"
+            python-version: "3.12"
           - ansible-version: "milestone"
             python-version: "3.12"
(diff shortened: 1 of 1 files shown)


oraNod:
@yurnov Hey, thanks so much for creating this pull request!

As you mentioned in email: "On a normal PR the splitter only runs the targets touched by the diff, so the extra leg costs one or two jobs. With the test-all-the-targets label the full fan-out goes from 24 to 32 jobs, and each one builds a kind cluster, so it is roughly a third more on the heaviest path."

This is a good point and worth some discussion before merging. I see that the integration tests against 2.21 are passing and this closes the gap for testing against the core version that pairs with the current Ansible community package version. While that's a good thing, maybe it's a bit much for the integration tests. Unit and sanity tests are already running against 2.21, which might be adequate. Integration tests run against milestone and are a good indicator since devel is usually a version ahead of the current package version.

Not for this PR but I do think that we could make the integration test workflows a little more efficient if we do decide to merge these changes. For instance we could cache Python dependencies and do things like building dependency collections in a separate job and then share via artifacts. However I don't think we can do much for the need to set up kind in each run.

Thanks again for your contribution with this PR! I've added some other reviewers and will let them share thoughts on all the above.

yurnov:
Hi @oraNod, 

Thanks for your comment, hope that @abikouo and @beeankha will be able to participate in this discussion. 

The splitter in integration test runs only the necessary tests, and the full suite is triggered only when the tag `test-all-the-targets` is added (which is not optimal, as over time it hides that a test is broken by some upstream changes), so the effect on the average testing cost should not be significant. 

However, with `test-all-the-targets` it increases test runs from 24 to 32, that is my question. And for that, I consider that we can exclude the combination stable-2.18 / turbo off, which will keep the same number of tests as now. Branch stable-2.18 is chosen as the latest branch with support for turbo mode, as this version of ansible-core has reached EOL, so for me no reason to test 2.18 without turbo mode. On the other hand, the number of PR (and CI runs) is low here, so additional runs should not significantly increase costs, and it was a reason why I keep it in the matrix.

Let me know your opinion if you decide to remove stable-2.18 / turbo off; I will exclude it as a separate commit. 

And following up on #1224 and my comments above regarding the silently broken test, I propose considering a scheduled (i.e., bi-weekly) CI run of the integration test suite.

oraNod:
> Let me know your opinion if you decide to remove stable-2.18 / turbo off; I will exclude it as a separate commit.

@yurnov I've spoken to @GomathiselviS about this one and agree that it makes sense to drop stable-2.18 / turbo off. If you could send a commit for that it would be great. If not, please let me know and I'd be happy to follow up. Thanks!

oraNod (approved):
LGTM. Thanks @yurnov

yurnov:
Funny part, that when I started to exclude 2.18/turbo off, I noticed a typo in the branch name, `text/add-ansible-2.21` (`text` instead of `test`)

abikouo (approved):

beeankha (approved):

PR #1236, opened 2026-09-08
fix(helm_repository): normalize both sides of the repo URL comparison

### LINKED ISSUE #1235: helm_repository: fails with "Repository already have a repository named X" when the repo is registered with a trailing slash
##### SUMMARY
`helm_repository` fails with `Repository already have a repository named <name>` when the repository is already registered with a trailing slash in its URL and the task passes that same URL, trailing slash included. The task is a plain no-op case, so it should report `changed=false`.

This is a regression from #1121, reported by @RobinR1 in https://github.com/ansible-collections/kubernetes.core/issues/480#issuecomment-5575892797.

That PR normalizes the URL coming from the task:

```python
repo_url = module.params.get("repo_url")
if repo_url:
    repo_url = repo_url.rstrip("/")          # plugins/modules/helm_repository.py:306
```

but the value it is later compared against comes straight from `helm repo list --output=yaml`, and helm stores (and returns) the URL verbatim, trailing slash included:

```python
elif repository_status["url"] != repo_url:   # plugins/modules/helm_repository.py:334
    module.fail_json(msg="Repository already have a repository named {0}".format(repo_name))
```

The comparison is therefore asymmetric: only one side is normalized. PR #1121 fixed the "task URL has a trailing slash, stored URL does not" direction and broke the mirror case, "stored URL has a trailing slash, task URL has one too".

Normalizing both sides of the comparison is enough to fix it, and it keeps the behaviour PR #1121 was after:

```python
elif repository_status["url"].rstrip("/") != repo_url:
```

Workaround in the meantime: `force: true` (alias of `force_update`) is evaluated before the URL comparison, so it bypasses the failure, at the cost of idempotency, since that branch unconditionally sets `changed=true` and re-runs `helm repo add` on every play.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

helm_repository

##### ANSIBLE VERSION
Not related; tested with ansible-core 2.18.19

##### COLLECTION VERSION
PR #1121 was backported to `stable-5` (#1125) and `stable-6` (#1126), so 5.4.3 is affected as well: 
**stable-5**: 5.4.3 and newer
**stable-6**: 6.5.0

##### CONFIGURATION
N/A

##### OS / ENVIRONMENT
N/A

##### STEPS TO REPRODUCE
```yaml
- hosts: localhost
  connection: local
  gather_facts: false
  tasks:
    - name: Register the repo with a trailing slash (helm stores the URL verbatim)
      ansible.builtin.command: helm repo add cilium https://helm.cilium.io/

    - name: Same name, same URL - expected to be an idempotent no-op
      kubernetes.core.helm_repository:
        name: cilium
        repo_url: https://helm.cilium.io/
```

##### EXPECTED RESULTS

The second task is a no-op and returns `changed=false`, since the repository is already registered under that name with exactly the URL requested.

##### ACTUAL RESULTS

```paste below
TASK [Register the repo with a trailing slash (helm stores the URL verbatim)] ***
changed: [localhost]

TASK [Same name, same URL - expected to be an idempotent no-op] ****************
    "changed": false,
    "invocation": {
        "module_args": {
            "api_key": null,
            "binary_path": null,
            "ca_cert": null,
            "context": null,
            "force_update": false,
            "host": null,
            "insecure_skip_tls_verify": false,
            "kubeconfig": null,
            "name": "cilium",
            "pass_credentials": false,
            "repo_name": "cilium",
            "repo_password": null,
            "repo_state": "present",
            "repo_url": "https://helm.cilium.io/",
            "repo_username": null,
            "validate_certs": true
        }
    },
    "msg": "Repository already have a repository named cilium"
}

PLAY RECAP *********************************************************************
localhost                  : ok=1    changed=1    unreachable=0    failed=1    skipped=0    rescued=0    ignored=0
```

For reference, this is what helm has on disk at that point:

```
helm repo list --output=yaml
- name: cilium
  url: https://helm.cilium.io/
```

yurnov:
##### SUMMARY
helm stores repository URLs verbatim, so a repository registered with a trailing slash keeps it in `repositories.yaml`.  #1121 normalized only the URL coming from the task, leaving the comparison asymmetric, so a repository that already matched was reported as a name conflict. Normalize both sides.

Fixes #1235

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
helm_repository

##### ADDITIONAL INFORMATION
The integration target only covered "stored without slash / requested without slash". Added the three remaining combinations, plus a negative case asserting a genuinely different URL still fails. `test_helm_repo_slash` is registered by calling `helm repo add` directly, since the module can no longer produce a stored URL with a trailing slash.

### DIFF
--- a/changelogs/fragments/20260908-helm-repository-trailing-slash.yaml
+++ b/changelogs/fragments/20260908-helm-repository-trailing-slash.yaml
@@ -1,0 +1,3 @@
+---
+bugfixes:
+  - helm_repository - normalize both sides of the repository URL comparison, so that a repository already registered with a trailing slash is recognized as matching instead of failing with ``Repository already have a repository named <name>`` (https://github.com/ansible-collections/kubernetes.core/pull/1236).
--- a/plugins/modules/helm_repository.py
+++ b/plugins/modules/helm_repository.py
@@ -331,7 +331,7 @@ def main():
                 insecure_skip_tls_verify,
             )
             changed = True
-        elif repository_status["url"] != repo_url:
+        elif repository_status["url"].rstrip("/") != repo_url:
             module.fail_json(
                 msg="Repository already have a repository named {0}".format(repo_name)
             )
--- a/tests/integration/targets/helm_repository/tasks/run_tests.yml
+++ b/tests/integration/targets/helm_repository/tasks/run_tests.yml
@@ -103,3 +103,81 @@
     binary_path: "{{ helm_binary }}"
     name: test_helm_repo
     state: absent
+
+# https://github.com/ansible-collections/kubernetes.core/issues/1235
+- name: Ensure the trailing slash test repositories don't exist
+  helm_repository:
+    binary_path: "{{ helm_binary }}"
+    name: "{{ item }}"
+    state: absent
+  loop:
+    - test_helm_repo_slash
+    - test_helm_repo_noslash
+
+- name: Register test_helm_repo_slash directly through helm, keeping the trailing slash
+  ansible.builtin.command: "{{ helm_binary }} repo add test_helm_repo_slash {{ chart_test_repo }}/"
+  changed_when: true
+
+- name: Add test_helm_repo_noslash chart repository without a trailing slash
+  helm_repository:
+    binary_path: "{{ helm_binary }}"
+    name: test_helm_repo_noslash
+    repo_url: "{{ chart_test_repo }}"
+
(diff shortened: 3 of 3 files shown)


beeankha (approved):

oraNod (approved):

oraNod:
Thanks @yurnov and @mandar242

PR #1120, opened 2026-05-06
docs: clarify ansible turbo mode with version compatibility notice

yurnov:
##### SUMMARY
The `docs/ansible_turbo_mode.rst` file was accidentally removed in PR #988, breaking the link referenced in `README.md`. Restore it so users following that link get the full Turbo mode documentation rather than a 404.

Additionally, document the `ansible-core` version compatibility issue for `cloud.common` (see ansible-collections/cloud.common@5e3430d and kubernetes.core PR #1109): Turbo mode is incompatible with `ansible-core>=2.19` because `cloud.common` no longer supports it. Without this clarification, users on newer ansible-core versions encounter silent or confusing failures when `ENABLE_TURBO_MODE=1` is set.

Update `README.md` with the same warning so the incompatibility is visible before users follow the link to the full docs.

##### ISSUE TYPE

- Docs Pull Request

##### COMPONENT NAME
documentation

##### ADDITIONAL INFORMATION
Addresses: #927

PR #1242, opened 2026-09-11
Add a deprecation notice for ENABLE_TURBO_MODE

mandar242:
### SUMMARY

Emit an explicit deprecation warning when users enable Ansible Turbo mode via `ENABLE_TURBO_MODE`, and stop turbo mode from being used on ansible-core versions where `cloud.common` is not supported.

Turbo mode depends on `cloud.common`, which is being retired. PR #1120 already deprecated the `cloud.common` dependency; without this change, removing that dependency later would cause `ENABLE_TURBO_MODE` to stop working silently. This PR warns users at runtime so they can migrate before support is removed in `kubernetes.core` 8.0.0.

Changes:
- Modules: emit the deprecation at import time via the module-level `deprecate()` in `plugins/module_utils/ansiblemodule.py`. `AnsibleTurboModule.__init__()` ends in `run_on_daemon()`, which calls `exit_json()` and exits the process, so a `deprecate()` call tied to instance construction is never reached.
- Modules: skip turbo mode on ansible-core >= 2.19.0 and fall back to the standard `AnsibleModule`, emitting a warning. `cloud.common` declares `requires_ansible '>=2.15.0,<2.19'`, but that is only enforced at collection load time and the turbo import itself still succeeds on newer cores, so turbo mode would otherwise fail later with `byte indices must be integers or slices, not str`.
- Modules and lookup: warn instead of silently falling back when `ENABLE_TURBO_MODE` is set but `cloud.common` is not installed.
- Lookup: call `Display.deprecated()` in the `k8s` lookup when turbo mode is enabled. The lookup path still works on >= 2.19, so it keeps using turbo and is not version-gated.
- Docs: add deprecation notices to `README.md` and `docs/ansible_turbo_mode.rst`.
- Changelog: add `deprecated_features`, `bugfixes`, and `minor_changes` fragment entries.

On ansible-core < 2.19 with `cloud.common` installed, turbo mode continues to work exactly as before and only gains a deprecation warning. On ansible-core >= 2.19 turbo mode is now deliberately bypassed, which replaces a confusing traceback with a clear warning.

### ISSUE TYPE

- Bugfix Pull Request
- Docs Pull Request

### COMPONENT NAME

`ENABLE_TURBO_MODE` / Ansible Turbo mode

### ADDITIONAL INFORMATION

On a supported ansible-core (< 2.19) with `cloud.common` installed, turbo mode stays active and users see:

```
[DEPRECATION WARNING]: ENABLE_TURBO_MODE is deprecated, as it relies on the cloud.common collection which is being retired. Setting this environment variable will have no effect once support is removed. This feature will be removed from kubernetes.core in version 8.0.0.
```

On ansible-core >= 2.19.0, turbo mode is bypassed and an additional warning is emitted:

```
[WARNING]: ENABLE_TURBO_MODE is ignored on ansible-core 2.21.4: Ansible Turbo mode requires the cloud.common collection, which supports ansible-core < 2.19.0 only. Continuing without Turbo mode.
```

Warnings are emitted from:
- `plugins/module_utils/ansiblemodule.py` for collection modules
- `plugins/lookup/k8s.py` for the `k8s` lookup plugin

To verify locally:

```bash
export ENABLE_TURBO_MODE=1
ansible localhost -m kubernetes.core.k8s_info -a "kind=Namespace" 2>&1 | grep -iE "deprecat|warning"
```

Integration tests that run with `ENABLE_TURBO_MODE=true` use ansible-core 2.18, where turbo mode remains enabled, so turbo behaviour there is unchanged.

Thanks to @yurnov for diagnosing the unreachable deprecation and for the suggested fix in mandar242/kubernetes.core#1.

### DIFF
--- a/README.md
+++ b/README.md
@@ -181,6 +181,8 @@ For documentation on how to use individual modules and other content included in
 
 ## Ansible Turbo Mode Tech Preview
 
+> **Deprecated:** Ansible Turbo mode (`ENABLE_TURBO_MODE`) is deprecated and will be removed in a future major release (planned for 8.0.0), since it depends on the `cloud.common` collection, which is being retired. Setting `ENABLE_TURBO_MODE` will keep emitting a deprecation warning until support is removed, at which point the variable will have no effect.
+
 > **Note:** Ansible Turbo mode is supported only with ``ansible-core`` versions **lower than 2.19**. Starting from ``ansible-core`` 2.19, the ``cloud.common`` collection is no longer supported, and therefore Ansible Turbo mode is also not supported in ``kubernetes.core``. If ``ENABLE_TURBO_MODE`` is set to ``1`` (or any truthy value) with ``ansible-core >= 2.19.0``, it may lead to fails.
 
 The ``kubernetes.core`` collection supports Ansible Turbo mode as a tech preview via the ``cloud.common`` collection (requires `ansible-core < 2.19`). By default, this feature is disabled. To enable Turbo mode for modules, set the environment variable `ENABLE_TURBO_MODE=1` on the managed node. For example:
--- a/plugins/module_utils/ansiblemodule.py
+++ b/plugins/module_utils/ansiblemodule.py
@@ -14,12 +14,24 @@ except TypeError:
 
 if enable_turbo_mode:
     try:
-        from ansible_collections.cloud.common.plugins.module_utils.turbo.module import (  # noqa: F401
-            AnsibleTurboModule as AnsibleModule,
+        from ansible_collections.cloud.common.plugins.module_utils.turbo.module import (
+            AnsibleTurboModule as _BaseAnsibleModule,
         )
 
-        AnsibleModule.collection_name = "kubernetes.core"
+        _BaseAnsibleModule.collection_name = "kubernetes.core"
     except ImportError:
-        from ansible.module_utils.basic import AnsibleModule  # noqa: F401
+        from ansible.module_utils.basic import AnsibleModule as _BaseAnsibleModule
+
+    class AnsibleModule(_BaseAnsibleModule):
+        def __init__(self, *args, **kwargs):
+            super(AnsibleModule, self).__init__(*args, **kwargs)
+            self.deprecate(
+                "ENABLE_TURBO_MODE is deprecated, as it relies on the cloud.common "
+                "collection which is being retired. Setting this environment "
+                "variable will have no effect once support is removed.",
+                version="8.0.0",
+                collection_name="kubernetes.core",
+            )
+
 else:
     from ansible.module_utils.basic import AnsibleModule  # noqa: F401
--- a/changelogs/fragments/deprecate-enable-turbo-mode.yml
+++ b/changelogs/fragments/deprecate-enable-turbo-mode.yml
@@ -1,4 +1,4 @@
 deprecated_features:
   - Ansible Turbo mode (``ENABLE_TURBO_MODE``) has been deprecated and will be removed
     in release 8.0.0, as it depends on the ``cloud.common`` collection, which is being
-    retired (https://github.com/ansible-collections/kubernetes.core/pull/1240).
+    retired (https://github.com/ansible-collections/kubernetes.core/pull/1242).
--- a/changelogs/fragments/deprecate-enable-turbo-mode.yml
+++ b/changelogs/fragments/deprecate-enable-turbo-mode.yml
@@ -1,0 +1,4 @@
+deprecated_features:
+  - Ansible Turbo mode (``ENABLE_TURBO_MODE``) has been deprecated and will be removed
+    in release 8.0.0, as it depends on the ``cloud.common`` collection, which is being
+    retired (https://github.com/ansible-collections/kubernetes.core/pull/1240).
--- a/docs/ansible_turbo_mode.rst
+++ b/docs/ansible_turbo_mode.rst
@@ -13,6 +13,13 @@ Following document provides overview of Ansible Turbo mode in ``kubernetes.core`
 
 
 .. warning::
+   Ansible Turbo mode is **deprecated** and will be removed in a future major release
+   (planned for 8.0.0), as it depends on the ``cloud.common`` collection, which is being
+   retired. Setting ``ENABLE_TURBO_MODE`` now emits a deprecation warning; once support is
+   removed, the variable will have no effect. Start planning your migration away from
+   Turbo mode.
+
+.. warning::
    Ansible Turbo mode is supported only with ``ansible-core`` versions **lower than 2.19**.
    Starting from ``ansible-core`` 2.19, the ``cloud.common`` collection is no longer supported,
    and therefore Ansible Turbo mode is also not supported in ``kubernetes.core``.
--- a/plugins/lookup/k8s.py
+++ b/plugins/lookup/k8s.py
@@ -213,6 +213,7 @@ from collections.abc import KeysView
 
 from ansible.errors import AnsibleError
 from ansible.module_utils.common.validation import check_type_bool
+from ansible.utils.display import Display
 from ansible_collections.kubernetes.core.plugins.module_utils.k8s.client import (
     get_api_client,
 )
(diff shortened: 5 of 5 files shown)


plugins/module_utils/ansiblemodule.py
```diff
@@ -14,12 +14,24 @@
...
+            super(AnsibleModule, self).__init__(*args, **kwargs)
+            self.deprecate(
+                "ENABLE_TURBO_MODE is deprecated, as it relies on the cloud.common "
+                "collection which is being retired. Setting this environment "
+                "variable will have no effect once support is removed.",
+                version="8.0.0",
+                collection_name="kubernetes.core",
+            )
+
 else:
```
yurnov, line 54: The deprecation never reaches the user.

`AnsibleTurboModule.__init__` ends with `run_on_daemon()`, which calls `exit_json()` and exits the process. So in the subclass, `super().__init__()` never returns and the `self.deprecate()` call after it is unreachable.

Tested with `ENABLE_TURBO_MODE=1`, `k8s_info kind=Namespace`, cloud.common 5.0.0 and ansible-core 2.18.19, and with 2.19.13 (failed with `byte indices must be integers or slices, not str`), in both cases deprecation wasn't thrown. 

Moreover, the import succeeds on 2.19.13, so `except ImportError` covers only the case where cloud.common is missing.

Suggested change that will trow deprecatiom msg on Ansible < 2.19 and disable turbo mode from 2.19 onwards, where turbo mode is broken in any case

```suggestion
from ansible.module_utils.ansible_release import __version__ as ansible_version
from ansible.module_utils.common.validation import check_type_bool
from ansible.module_utils.common.warnings import deprecate
from ansible.module_utils.common.warnings import warn
from ansible_collections.kubernetes.core.plugins.module_utils.version import (
    LooseVersion,
)

# cloud.common declares requires_ansible '>=2.15.0,<2.19'. On newer ansible-core the
# import still succeeds, so the version has to be checked explicitly: relying on
# ImportError alone lets turbo mode fail later with an unrelated traceback.
TURBO_UNSUPPORTED_CORE = "2.19.0"

try:
    enable_turbo_mode = check_type_bool(os.environ.get("ENABLE_TURBO_MODE"))
except TypeError:
    enable_turbo_mode = False

if enable_turbo_mode:
    deprecate(
        "ENABLE_TURBO_MODE is deprecated, as it relies on the cloud.common collection "
        "which is being retired. Setting this environment variable will have no effect "
        "once support is removed.",
        version="8.0.0",
        collection_name="kubernetes.core",
    )
    if LooseVersion(ansible_version) >= LooseVersion(TURBO_UNSUPPORTED_CORE):
        warn(
            "ENABLE_TURBO_MODE is ignored on ansible-core %s: Ansible Turbo mode requires "
            "the cloud.common collection, which supports ansible-core < %s only. "
            "Continuing without Turbo mode." % (ansible_version, TURBO_UNSUPPORTED_CORE)
        )
        enable_turbo_mode = False

if enable_turbo_mode:
    try:
        from ansible_collections.cloud.common.plugins.module_utils.turbo.module import (  # noqa: F401
            AnsibleTurboModule as AnsibleModule,
        )

        AnsibleModule.collection_name = "kubernetes.core"
    except ImportError:
        warn(
            "ENABLE_TURBO_MODE is set but the cloud.common collection is not installed. "
            "Continuing without Turbo mode."
        )
        from ansible.module_utils.basic import AnsibleModule  # noqa: F401
else:
```

Tested with cloud.common 5.0.0 and the latest stable ansible (2.21.4):
```
[WARNING]: No inventory was parsed, only implicit localhost is available
[WARNING]: Collection cloud.common does not support Ansible version 2.21.4
[WARNING]: ENABLE_TURBO_MODE is ignored on ansible-core 2.21.4: Ansible Turbo mode requires the cloud.common collection, which supports ansible-core < 2.19.0 only. Continuing without Turbo mode.
[WARNING]: Deprecation warnings can be disabled by setting `deprecation_warnings=False` in ansible.cfg.
[DEPRECATION WARNING]: ENABLE_TURBO_MODE is deprecated, as it relies on the cloud.common collection which is being retired. Setting this environment variable will have no effect once support is removed. This feature will be removed from collection 'kubernetes.core' version 8.0.0.
localhost | SUCCESS => {
    "api_found": true,
...
```
Worth noting that I installed version 5.0.0 manually, `ansible-galaxy collection install cloud.common` on 2.21.4 resolves to 4.2.0, not 5.0.0. Reason of that is 4.2.0 declares `requires_ansible: '>=2.15.0'` with no upper cap, so the resolver picks it once 5.0.0 is excluded due to `requires_ansible: '>=2.15.0,<2.19'`

And with ansible 2.18.19:
```
[WARNING]: No inventory was parsed, only implicit localhost is available
[DEPRECATION WARNING]: ENABLE_TURBO_MODE is deprecated, as it relies on the 
cloud.common collection which is being retired. Setting this environment 
variable will have no effect once support is removed. This feature will be 
removed from kubernetes.core in version 8.0.0. Deprecation warnings can be 
disabled by setting deprecation_warnings=False in ansible.cfg.
localhost | SUCCESS => {
    "api_found": true,
    "changed": false,
```

yurnov (changes requested):

plugins/module_utils/ansiblemodule.py
yurnov: For some reason, the formatting shown suggested a change only to line R36; it should be range R8-R36

yurnov:
Hi @mandar242,

Please take a look at https://github.com/mandar242/kubernetes.core/pull/1. The general idea is to fix the warning and also fall back to turbo mode being disabled for unsupported versions of Ansible

README.md
```diff
@@ -181,6 +181,8 @@ For documentation on how to use individual modules and other content included in
 
 ## Ansible Turbo Mode Tech Preview
 
+> **Deprecated:** Ansible Turbo mode (`ENABLE_TURBO_MODE`) is deprecated and will be removed in a future major release (planned for 8.0.0), since it depends on the `cloud.common` collection, which is being retired. Setting `ENABLE_TURBO_MODE` will keep emitting a deprecation warning until support is removed, at which point the variable will have no effect.
+
 > **Note:** Ansible Turbo mode is supported only with ``ansible-core`` versions **lower than 2.19**. Starting from ``ansible-core`` 2.19, the ``cloud.common`` collection is no longer supported, and therefore Ansible Turbo mode is also not supported in ``kubernetes.core``. If ``ENABLE_TURBO_MODE`` is set to ``1`` (or any truthy value) with ``ansible-core >= 2.19.0``, it may lead to fails.
```
yurnov:
```suggestion
> **Note:** Ansible Turbo mode is supported only with ``ansible-core`` versions **lower than 2.19**. Starting from ``ansible-core`` 2.19, the ``cloud.common`` collection is no longer supported, and therefore Ansible Turbo mode is also not supported in ``kubernetes.core``. If ``ENABLE_TURBO_MODE`` is set to ``1`` (or any truthy value) with ``ansible-core >= 2.19.0``, Turbo mode is ignored and a warning is emitted; modules run without it.
```

yurnov: One final comment

yurnov (commented):
One final comment,

plugins/module_utils/ansiblemodule.py
mandar242: Good catch, thank you — you are right that the deprecation was unreachable. `run_on_daemon()` exits the process inside `AnsibleTurboModule.__init__()`, so nothing after `super().__init__()` ever runs.

Applied your suggestion in fe32680. The deprecation now fires at import time via the module-level `deprecate()`, which lands in the global warnings registry that `_return_formatted()` drains on `exit_json()`, so it survives the early exit. Also added the explicit `LooseVersion` gate on `TURBO_UNSUPPORTED_CORE = "2.19.0"` and the `warn()` on the `ImportError` path.

Verified locally across all four paths: core < 2.19 with cloud.common installed still resolves to `AnsibleTurboModule`; core >= 2.19 falls back to the stock `AnsibleModule` with the warning; cloud.common missing falls back with its own warning; turbo disabled is unchanged. Changelog fragment now carries `bugfixes` and `minor_changes` entries alongside the deprecation.

README.md
mandar242: Applied in fcbc2ca — thanks. I also updated the matching sentence in `docs/ansible_turbo_mode.rst`, which carried the same "may lead to failures" wording, so the two stay consistent.

yurnov (approved):
LGTM

oraNod (approved):
🎉 Thanks @mandar242 and @yurnov

chynasan (approved):

yurnov:
> ### Backport to stable-5: 💔 cherry-picking failed — conflicts found
> ❌ Failed to cleanly apply [031bc47](https://github.com/ansible-collections/kubernetes.core/commit/031bc475fadd0fcf8152d3c9083e22a77e3d870d) on top of patchback/backports/stable-5/031bc475fadd0fcf8152d3c9083e22a77e3d870d/pr-1242
> 

Manual backport created as #1251

PR #1251, opened 2026-09-14
[PR #1242/031bc475 backport][stable-5] Add a deprecation notice for ENABLE_TURBO_MODE

yurnov:
**This is a manual backport of PR #1242 as merged into main (031bc475fadd0fcf8152d3c9083e22a77e3d870d).**

### SUMMARY

Emit an explicit deprecation warning when users enable Ansible Turbo mode via `ENABLE_TURBO_MODE`, and stop turbo mode from being used on ansible-core versions where `cloud.common` is not supported.

Turbo mode depends on `cloud.common`, which is being retired. PR #1120 already deprecated the `cloud.common` dependency; without this change, removing that dependency later would cause `ENABLE_TURBO_MODE` to stop working silently. This PR warns users at runtime so they can migrate before support is removed in `kubernetes.core` 8.0.0.

Changes:
- Modules: emit the deprecation at import time via the module-level `deprecate()` in `plugins/module_utils/ansiblemodule.py`. `AnsibleTurboModule.__init__()` ends in `run_on_daemon()`, which calls `exit_json()` and exits the process, so a `deprecate()` call tied to instance construction is never reached.
- Modules: skip turbo mode on ansible-core >= 2.19.0 and fall back to the standard `AnsibleModule`, emitting a warning. `cloud.common` declares `requires_ansible '>=2.15.0,<2.19'`, but that is only enforced at collection load time and the turbo import itself still succeeds on newer cores, so turbo mode would otherwise fail later with `byte indices must be integers or slices, not str`.
- Modules and lookup: warn instead of silently falling back when `ENABLE_TURBO_MODE` is set but `cloud.common` is not installed.
- Lookup: call `Display.deprecated()` in the `k8s` lookup when turbo mode is enabled. The lookup path still works on >= 2.19, so it keeps using turbo and is not version-gated.
- Docs: add deprecation notices to `README.md` and `docs/ansible_turbo_mode.rst`.
- Changelog: add `deprecated_features`, `bugfixes`, and `minor_changes` fragment entries.

On ansible-core < 2.19 with `cloud.common` installed, turbo mode continues to work exactly as before and only gains a deprecation warning. On ansible-core >= 2.19 turbo mode is now deliberately bypassed, which replaces a confusing traceback with a clear warning.

### ISSUE TYPE

- Bugfix Pull Request
- Docs Pull Request

### COMPONENT NAME

`ENABLE_TURBO_MODE` / Ansible Turbo mode

### ADDITIONAL INFORMATION

On a supported ansible-core (< 2.19) with `cloud.common` installed, turbo mode stays active and users see:

```
[DEPRECATION WARNING]: ENABLE_TURBO_MODE is deprecated, as it relies on the cloud.common collection which is being retired. Setting this environment variable will have no effect once support is removed. This feature will be removed from kubernetes.core in version 8.0.0.
```

On ansible-core >= 2.19.0, turbo mode is bypassed and an additional warning is emitted:

```
[WARNING]: ENABLE_TURBO_MODE is ignored on ansible-core 2.21.4: Ansible Turbo mode requires the cloud.common collection, which supports ansible-core < 2.19.0 only. Continuing without Turbo mode.
```

Warnings are emitted from:
- `plugins/module_utils/ansiblemodule.py` for collection modules
- `plugins/lookup/k8s.py` for the `k8s` lookup plugin

To verify locally:

```bash
export ENABLE_TURBO_MODE=1
ansible localhost -m kubernetes.core.k8s_info -a "kind=Namespace" 2>&1 | grep -iE "deprecat|warning"
```

Integration tests that run with `ENABLE_TURBO_MODE=true` use ansible-core 2.18, where turbo mode remains enabled, so turbo behaviour there is unchanged.

Thanks to @yurnov for diagnosing the unreachable deprecation and for the suggested fix in mandar242/kubernetes.core#1.


mandar242 (approved):

changelogs/fragments/deprecate-enable-turbo-mode.yml
```diff
@@ -0,0 +1,7 @@
+---
+deprecated_features:
+  - Ansible Turbo mode (``ENABLE_TURBO_MODE``) has been deprecated and will be removed in release 8.0.0, as it depends on the ``cloud.common`` collection, which is being retired (https://github.com/ansible-collections/kubernetes.core/pull/1242).
+bugfixes:
+  - Ansible Turbo mode - ignore ``ENABLE_TURBO_MODE`` on ansible-core 2.19.0 and later, where the ``cloud.common`` collection is not supported, and fall back to the standard ``AnsibleModule`` instead of failing with ``byte indices must be integers or slices, not str`` (https://github.com/ansible-collections/kubernetes.core/pull/1242).
+minor_changes:
+  - k8s lookup - warn when ``ENABLE_TURBO_MODE`` is set but the ``cloud.common`` collection is not installed, instead of silently falling back to the standard lookup base (https://github.com/ansible-collections/kubernetes.core/pull/1242).
```
abikouo, line 7:
```suggestion
  - Ansible Turbo mode (``ENABLE_TURBO_MODE``) has been deprecated and will be removed in release 8.0.0, as it depends on the ``cloud.common`` collection, which is being retired (https://github.com/ansible-collections/kubernetes.core/pull/1251).
bugfixes:
  - Ansible Turbo mode - ignore ``ENABLE_TURBO_MODE`` on ansible-core 2.19.0 and later, where the ``cloud.common`` collection is not supported, and fall back to the standard ``AnsibleModule`` instead of failing with ``byte indices must be integers or slices, not str`` (https://github.com/ansible-collections/kubernetes.core/pull/1251).
minor_changes:
  - k8s lookup - warn when ``ENABLE_TURBO_MODE`` is set but the ``cloud.common`` collection is not installed, instead of silently falling back to the standard lookup base (https://github.com/ansible-collections/kubernetes.core/pull/1251).
```

plugins/lookup/k8s.py
```diff
@@ -213,24 +213,38 @@
...
 
+display = Display()
+
 try:
     enable_turbo_mode = check_type_bool(os.environ.get("ENABLE_TURBO_MODE"))
 except TypeError:
     enable_turbo_mode = False
 
 if enable_turbo_mode:
+    display.deprecated(
```
abikouo, line 232: We should also add a test against the ansible-core version. If it is greater than `2.19`, default to standard execution and raise a warning.

abikouo (dismissed):

changelogs/fragments/deprecate-enable-turbo-mode.yml
yurnov: Hi @abikouo, not valid as it not a normal PR, it's a cherry-pick of merged PR #1242 to stable-5 branch, as pachback failed to created such cererry-pick due to merge conflict.

So, it's just a manual cherry-pick

plugins/lookup/k8s.py
yurnov: valid comment, but I will prefer to add it as the separate PR rather than here, as it manual dropback (cherry-pick to `stable-5` branch), to keep consistency between main, stable-6 and stable-6

changelogs/fragments/deprecate-enable-turbo-mode.yml
abikouo: I just missed the fact that this is a backport PR

plugins/lookup/k8s.py
oraNod: @abikouo and @yurnov I've opened a PR for the test: https://github.com/ansible-collections/kubernetes.core/pull/1253

Although I could be misinterpreting the details with that. I'd appreciate review. Thanks!

oraNod (approved):

